PR #149, opened 2026-07-12
feat: underwater seabed reflection, ambient noise and ship-traffic source levels

jmrplens:
## Summary

Completes the closed-form **underwater sound propagation** sub-phase (P1),
adding three modules on top of the transmission-loss / sound-speed / sonar core:

- **`seabed_reflection`** — fluid–fluid **Rayleigh** reflection coefficient
  (Medwin & Clay), the **critical grazing angle** `arccos(c1/c2)` (total
  reflection below it) and the **bottom loss** `BL = −20·lg|R|`.
- **`ocean_ambient_noise`** — **Wenz-framework** spectrum levels, energy-summing
  **wind noise** (the "rule of fives", 25 dB @ 1 kHz / 5 kn) and **Mellen
  thermal noise** (`4π·k·T·ρ·f²/c`), with an optional caller-supplied shipping
  term.
- **`ship_traffic_noise`** — predicted **ship source-level spectra** from vessel
  class, speed and length. Three semi-empirical models: **JOMOPANS-ECHO**
  (MacGillivray & de Jong 2021, per vessel class, the default), **RANDI 3.1** and
  **Wales & Heitmeyer** (2002).

## Clean-room provenance

Implemented from the equations, not from any open-source code. The
JOMOPANS-ECHO paper is open access (CC-BY); the implementation is validated
against the **authors' own reference calculator** (supplementary File S1, an
Excel workbook) to **better than 0.01 dB** across the full decidecade band —
this also resolved an OCR ambiguity in the paper's `f̂1` term (`480/V_C` general,
`600/V_C` for the cargo low-frequency hump). Seabed reflection is checked against
the normal-incidence impedance formula; ambient noise against the published
rule-of-fives anchor and the Mellen formula.

## Why the shipping model

The ambient-noise module leaves the low-frequency turbulence band and a built-in
distant-shipping curve out of scope (Wenz notes these are strongly variable). To
still give users a *default* option when they have no measured spectrum, the
`ship_traffic_noise` model predicts the shipping term from readily available
traffic parameters and can be fed straight into `ocean_ambient_noise(...,
shipping=...)`.

## Each module ships with

- A frozen result dataclass with `.plot()`.
- A conformance check backed by an **independent** oracle (total **179/179**).
- One figure in four language × theme variants (12 new SVGs).
- English docs (`docs/underwater-propagation.md`) plus EN/ES site guides,
  api-reference, indices and the top-level README.

## Verification

- `ruff check .` — clean
[...]

PR #150, opened 2026-07-12
ci(docs): install latest pnpm via corepack to fix docs deploy

jmrplens:
## Problem

The **Deploy Docs** workflow has been failing at the `pnpm/action-setup@v6`
step, so the documentation never deploys. With the floating `version: 11`, the
action installs a bundled pnpm and then self-updates to the latest 11.x; that
switch crashes:

```
Switching pnpm from v11.7.0 to v11.12.0...
[ERROR] Cannot use 'in' operator to search for 'integrity' in undefined
    at createFullPkgId (…/setup-pnpm/node_modules/pnpm/dist/pnpm.mjs…)
    at lockfileToDepGraph
```

The build job dies here, before the site build / `html-validate` / `pa11y`
steps ever run. (Two `long-title` html-validate errors that would have failed
the build once setup was fixed were resolved separately in #149.)

## Fix

Install pnpm with **corepack** (shipped with Node) instead of
`pnpm/action-setup`. `corepack prepare pnpm@latest --activate` installs the
latest pnpm cleanly and always tracks the newest release, and the pnpm store is
cached with `actions/cache`.

The current `pnpm-lock.yaml` (lockfile v9.0) is read fine by the latest pnpm, so
no lockfile change is needed.

## Verification (local, pnpm 11.12.0)

- `pnpm install --frozen-lockfile` — OK
- `pnpm build` — 95 pages built
- `pnpm run html-validate` — exit 0
- `pnpm run pa11y` — 10/10 URLs, 0 errors

The Deploy Docs workflow only runs on push to `main`, so it will be exercised
for real once this merges.

## Summary by Sourcery

Update docs deployment workflow to install pnpm via corepack and cache its store, replacing the failing pnpm/action-setup step.

CI:
- Replace pnpm/action-setup with corepack-based pnpm installation in the docs workflow.
- Add explicit pnpm store caching using actions/cache keyed by the docs pnpm lockfile.

## Summary by CodeRabbit

* **Chores**
  * Improved the documentation build workflow by updating package management setup and caching.
  * Documentation builds now use the latest pnpm release and benefit from faster dependency installation.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,14 +22,23 @@ jobs:
           persist-credentials: false
           # Full history: add-sitemap-lastmod.mjs stamps per-file git dates.
           fetch-depth: 0
-      - uses: pnpm/action-setup@v6
-        with:
-          version: 11
       - uses: actions/setup-node@v6
         with:
           node-version: 22
-          cache: pnpm
-          cache-dependency-path: site/pnpm-lock.yaml
+      - name: Enable latest pnpm via corepack
+        # corepack ships with Node and installs pnpm cleanly. We avoid
+        # pnpm/action-setup, whose bundled installer crashes while self-updating
+        # to the latest release ("Cannot use 'in' operator ... integrity").
+        run: |
+          corepack enable
+          corepack prepare pnpm@latest --activate
+          pnpm --version
+      - name: Cache pnpm store
+        uses: actions/cache@v4
+        with:
+          path: ~/.local/share/pnpm/store
+          key: pnpm-store-${{ runner.os }}-${{ hashFiles('site/pnpm-lock.yaml') }}
+          restore-keys: pnpm-store-${{ runner.os }}-
       - name: Regenerate llms.txt from docs (freshness guarantee)
         run: python3 scripts/generate_llms.py
       - name: Check EN/ES translation parity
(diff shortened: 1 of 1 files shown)


jmrplens:
Thanks. Addressed in 56bfa70:

- **pnpm store path (copilot, coderabbit):** now derived from `pnpm store path`
  instead of the hardcoded `~/.local/share/pnpm/store`. Verified locally it
  resolves to `.../store/v11`, so the old hardcoded key would indeed have
  missed. Good catch.
- **actions/cache version:** bumped v4 → v6 (latest major).

Kept as-is, by design:

- **`pnpm@latest` determinism (copilot, coderabbit, sourcery):** this is a
  deliberate project decision — the workflow must always track the newest pnpm
  release, not pin a major/exact version. `--frozen-lockfile` still guarantees
  reproducible dependency resolution regardless of the pnpm binary version, so
  the trade-off is limited to the pnpm tool itself.
- **SHA-pinning actions/cache (coderabbit):** every action in this repo is
  referenced by major tag (checkout@v7, setup-node@v6, …); pinning one action to
  a commit SHA would be inconsistent and would freeze it off "latest", which is
  the opposite of the intent here. Kept the tag.

Verified locally with pnpm 11.12.0: install --frozen-lockfile, build,
html-validate (0 errors) and pa11y (10/10, 0 errors) all pass.

PR #161, opened 2026-07-13
chore: golden baseline and hotspot benchmarks (overhaul phase 0)

jmrplens:
## Summary

Phase 0 of the library overhaul: a deterministic golden baseline plus micro-benchmarks, so the upcoming package modularization and the vectorization sprints can prove numeric neutrality.

- **`scripts/bench.py`**: fixed synthetic inputs for the hotspot cases (Doc 29 contour chain, rotorcraft hemisphere nearest-bin fill, PE marching solver, ECMA-418-2 analysers, human-vibration Wd weighting), with three modes: benchmark table (default), `--figures` (per-figure wall time) and `--golden` (freeze outputs).
- **`tests/golden_data.py`**: 542 frozen values (auto-generated; regenerate only on reviewed numeric changes).
- **`tests/test_golden_baseline.py`**: exact-equivalence guard (rtol 1e-9; 1e-7 for the ECMA long filter chains, same rationale as the tolerance-based figure gate).

## Baseline: hotspots

| case | best | note |
|---|---|---|
| airport_contour_10x8 | 212 ms | tiny grid; real 60x48 grids scale ~36x |
| ecma_tonality_0.5s | 2.99 s | dominates figure generation |
| ecma_loudness_0.5s | 2.72 s | idem |
| ecma_roughness_0.5s | 250 ms | |
| rotorcraft_hemisphere_5q | 24 ms | 4.8 ms/query, multiplies into contours |
| underwater_pe_3km | 13 ms | |
| airport_event_level | 2.9 ms | |
| vibration_wd_weighting_10s | 0.96 ms | confirms drop from perf scope |

## Baseline: figure generation (single EN/light variant, 102 functions, 123.6 s)

| figure | wall |
|---|---|
| loudness_models_comparison | 36.0 s |
| tonality_roughness_demo | 22.4 s |
| airport_contour | 21.4 s |
| sottek_specific_loudness | 4.6 s |
| filter_responses | 3.0 s |
| numerical_propagation | 2.4 s |
| moore_glasberg_time_loudness | 2.2 s |
| weighting_responses | 1.7 s |
| sti_curve | 1.6 s |
| fluctuation_strength | 1.3 s |

No production code changes; the bench is not wired into CI.

[...]

PR #162, opened 2026-07-13
refactor: reorganize the package into twelve domain subpackages (3.2)

jmrplens:
## Summary

Phase 1 of the library overhaul: the 89 flat top-level modules move into **twelve domain subpackages** (plus the private `_internal` helpers and a per-domain `_plot` split of the 3.6k-line `_plotting.py`), with **zero user-visible breakage**:

- The **flat API is unchanged**: every `from phonometry import x` keeps working and remains the primary surface.
- Each domain is also a namespace: `from phonometry import aircraft as air; air.noise_contour(...)`.
- Every pre-3.2 **module path keeps importing** for one deprecation cycle through `sys.modules` shims (`phonometry/_compat.py`): silent import, `DeprecationWarning` on attribute access, removal in 4.0. `phonometry.hearing` and `phonometry.environmental` became real packages that silently re-export their old module surface.

| Subpackage | Modules |
|---|---|
| metrology | 8 (filters, weighting, levels, calibration, verifiers, GUM) |
| psychoacoustics | 13 | 
| hearing | 5 | 
| emission | 5 | 
| room | 6 | 
| materials | 8 | 
| building | 11 | 
| vibration | 4 | 
| environmental | 6 | 
| aircraft | 4 | 
| underwater | 10 | 
| electroacoustics | 2 |

## Mechanics

- Whole-file moves are pure `git mv` commits (history preserved); the move hashes are listed in `.git-blame-ignore-revs`.
- Tests mirror the tree (`tests/<domain>/`, 85 files); deep imports across tests/scripts point at the new canonical paths; `tests/test_deprecated_aliases.py` gains a frozen 85-path snapshot asserting every pre-move path still imports silently, plus a behavior test for every shim.
- New `tests/test_package_architecture.py`: ast-based cross-package edge whitelist (`_internal` stays leaf, six justified domain edges) plus fresh-interpreter smoke imports per subpackage.
- `_plot/` split keeps the lazy `.plot()` convention (renderers import result classes only under `TYPE_CHECKING`); `_plotting` remains as a silent re-export shim until 4.0.

## Verification

- Full suite **2647 passed** (including runs with 3.2-shim `DeprecationWarning` escalated to error: internal code never touches a shim).
- **Golden numeric baseline** (PR #161) byte-stable: the reorganization changes no numbers.
- Conformance **194/194**; ruff and mypy clean; figures untouched.
- `import phonometry` time: 1.09 s vs 1.17 s on main.
- Pickles created under 3.2 reference new module paths (one-line changelog note); 3.1 pickles load fine through the shims.

Release 3.2.0 follows after merge.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jmrplens:
Review response:

- **Architecture test loophole** (gemini): applied. Absolute self-imports (`from phonometry.x import y` inside the package) now fail the architecture test directly, so they cannot bypass the cross-package whitelist.
- **FutureWarning instead of DeprecationWarning** (gemini, high): *not applied*. The repository's established deprecation contract since 3.1 (`_warn_renamed`, the NEP 23 wording, every existing alias and its tests, and the CONTRIBUTING deprecation policy) uses `DeprecationWarning`, which is the standard category for developer-facing API renames; it is visible in `__main__` code and under pytest by design. `FutureWarning` is reserved for behavior changes that silently alter results for end users, which a module path rename is not. Switching categories mid-cycle would also make the 3.2 shims inconsistent with the already-released 3.1 aliases.

PR #166, opened 2026-07-13
perf: vectorise the airport noise contour grid

jmrplens:
ECAC Doc 29 noise_contour evaluated the grid with a per-point Python loop over the scalar segment kernel, O(grid x segments) interpreted calls; the 10x8 benchmark grid took ~212 ms and production-size grids were impractical.

The new _grid_event_levels kernel runs one vectorised pass per flight-path segment with every term broadcast over the observers: segment geometry, the section 4.5.5 nearest-end selection for maximum-level metrics and runway rolls, lateral attenuation (Eq. 4-18/4-19), engine-installation directivity (Eq. 4-15/4-16), duration (Eq. 4-14), finite-segment fraction (Eq. 4-20/4-21), start-of-roll directivity (Eq. 4-22..4-25) and the NPD lookups through a per-observer power/distance variant of the Eq. 4-3/4-4 arithmetic. The scalar kernel remains the implementation of event_level for single receivers.

The grid path mirrors the scalar loop expression for expression, so results are numerically identical:

- the frozen golden values (including the 10x8 contour case) pass unchanged at rtol 1e-9, and
- a new test compares every grid point against event_level across takeoff ground roll, landing rollout, bank angle, both metrics and both engine mountings, at 1e-9 absolute.

Measured: the benchmark 10x8 grid drops from ~212 ms to ~5 ms, and a 200x150 grid (30 000 points) evaluates in ~0.2 s.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/166?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   `EPNLResult.bandsharing_adjustment`, and a `start_band` parameter on
   `tone_correction` for the helicopter 50 Hz procedure.
 
+### Changed
+
+- ECAC Doc 29 `noise_contour` now evaluates the grid with one vectorised pass
+  per flight-path segment instead of a per-point Python loop; numerically
+  identical (guarded by the golden baseline and a scalar-equivalence test)
+  and ~45x faster on small grids, several hundred times on production-size
+  grids (30 000 points in ~0.2 s).
+
 ### Fixed
 
 - ECAC Doc 29: behind takeoff ground-roll segments the lateral attenuation and
--- a/src/phonometry/aircraft/airport_noise.py
+++ b/src/phonometry/aircraft/airport_noise.py
@@ -622,6 +622,200 @@ def _event_level_core(
     return total, seg_arr
 
 
+def _npd_level_grid(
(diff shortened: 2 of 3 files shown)


jmrplens:
Review follow-ups applied: the analysis-flagged cognitive complexity is resolved by splitting the kernel into per-term _grid_* helpers, each mirroring its scalar counterpart. From gemini's suggestions, three are in (broadcast roll masks, scalar zero directivity off takeoff segments, column-wise observer block); the dsor hoist for the psi computation is declined because psi must divide by the raw distance exactly as the scalar path does (the 1e-9 floor there applies only to the distance argument), and the two expressions differ for degenerate observers within 1e-9 m of the roll start, which would break the machine-precision equivalence contract. Performance unchanged (4.7 ms on the benchmark grid).

PR #167, opened 2026-07-14
fix: underwater Wenz reference pressure, solver guards and new independent oracles

jmrplens:
Standards-correctness follow-ups for the underwater domain (ambient noise, numerical propagation, sound speed, ship spectra), plus a layer of new independent validation. Every formula and constant of the domain was re-verified against the primary sources; the items below are what deviated.

## Fixed

- **Wenz wind noise was 26 dB low.** The Wenz/Knudsen rule-of-fives anchor ("25 dB at 1 kHz for 5 knots") is stated re 0.0002 dyn/cm2 = 20 uPa inside the quoted source, but the module returned the unconverted value labelled dB re 1 uPa: every wind spectrum level and the ambient composite were 26.02 dB low, and the wind/thermal crossover appeared at ~12 kHz instead of ~63 kHz. The anchor is now 25 + 20 lg(20) = 51.02 dB re 1 uPa2/Hz, which matches the published Wenz chart (~50 dB at 1 kHz for 4-6 kn); the slopes were verified unchanged. The original test and conformance anchors had inherited the same transcription and were not independent; they now pin the corrected value plus graphical Wenz-chart anchors. Ambient-noise figures regenerated (4 variants).
- **normal_modes resolution guards.** The default depth grid is now derived from the physics (max(400, ceil(60 D f / c_min)), capped at 20000), the eigensolve is restricted to the propagating band, eigenvalues inside the O(dz2) finite-difference error band about cutoff are discarded (previously a spurious extra propagating mode appeared at under-resolved settings: 67 instead of 66 modes at 500 Hz / 100 m with the old fixed default) and a warning fires when a retained near-cutoff mode needs a finer grid. Mode interpolation at the source/receiver depths is a single vectorised lookup.
- **ray_trace turning depths.** The sound-speed gradient is evaluated exactly per profile segment (piecewise-constant, as the piecewise-linear profile implies) instead of np.gradient on an interpolated fine grid, which smeared thermocline kinks: on a SOFAR-like kinked profile the turning-depth errors drop from 3.4/8.0 m to 0.2/0.0 m.
- **Documentation accuracy:** ISO 17208-2 uncertainty band edges (the standard's own list leaves 16-20 kHz unassigned; the conservative 4 dB reading is documented), the Ainslie-McColm ~10 % agreement claim (marginally exceeded at the extreme corners of its stated domain, a property of the published fit), the inherent Francois-Garrison A3 step at 20 C, the published validity domains of the three sound-speed equations (they extrapolate silently outside them), the 2.5-40 kn validity of the Wenz wind-doubling law, the paraxial (~15-20 deg) validity of the standard PE, and stale post-reorganisation cross-references. sound_speed_profile now evaluates in one vectorised pass.

## Added validation (all independent of the implementation)

- Image-source absolute transmission-loss anchors for the ideal waveguide: five converged image-sum values that pin the modal sum's absolute calibration (the previous analytic test shared the implementation's prefactor and could not catch a calibration error) and the long-range parabolic-equation field.
- The published UNESCO EOS-80 canonical check value (Fofonoff & Millard 1983: 1731.995 m/s), with the ITS-90 conversion.
- Six additional JOMOPANS File S1 rows (hump edges and high-frequency tail), verified against the authors' spreadsheet to the fourth decimal.
- The Ainslie-McColm Table I reference oceans swept 100 Hz-1 MHz against Francois-Garrison (max deviation 8.35 %).
- The RANDI 3.1 report's representative source levels (Breeding et al., NRL, Table 2): the ship-spectrum model reproduces the Large and Super Tanker rows to 0.06 dB at every tabulated frequency; the two printed cells that contradict the report's own equations, and the Fishing row whose assumed average is not reproducible, are excluded with the rationale documented in the test.

Conformance grows to 198 checks (UNESCO canonical value and the image-source anchor). Full suite green, ruff and mypy clean, figure regeneration byte-identical for the untouched figures.

Reference values are factual data points with attribution; no source text or code is redistributed.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected underwater wind-noise calibration (including crossover behavior) using the ISO reference.
  * Improved normal-mode accuracy near cutoff with more robust depth-grid handling and warnings for borderline modes.
  * Improved ray-tracing accuracy by applying sound-speed gradients per profile segment.
  * Improved sound-speed profile evaluation while preserving results.

* **Documentation**
  * Clarified ocean ambient noise “rule of fives” reference/units and validity range (including Spanish guide).

* **Tests**
  * Expanded underwater conformance and added new absolute calibration/oracle checks (including UNESCO sound-speed and ideal-waveguide TL comparisons), increasing coverage from 196 to 198.


jmrplens:
Review follow-ups applied: the four numerical-propagation figure variants are regenerated (the near-cutoff eigenvalue guard legitimately changes the modal curve: the mode it now discards sat inside the finite-difference error band); the analysis-flagged cognitive complexity is resolved by extracting the propagating-band filter into a helper; gemini's point about the vectorised profile path bypassing the salinity validation is fixed (explicit non-negative check, matching the scalar path, plus a depth check already present); CodeRabbit's indexing simplification and type hint are in.

PR #236, opened 2026-07-19
test(broadcast): document the EBU-derived loudness series and drop the CI audio fetch

jmrplens:
The only external download the CI performed was the EBU loudness test set (about 76 MB of authentic-programme audio). That audio cannot be redistributed: the EBU set is licensed for technical testing only and its cases 7/8 and 5/6 carry third-party programme rights, so it was never committed.

What runs where is now made explicit and network-free in CI:

- `tests/data/broadcast/README.md` documents the committed per-block loudness series (`ebu_programme_block_loudness.npz`): their EBU Tech 3341/3342 origin, that they are derived measurement data (a 100 ms-hop loudness envelope cannot reconstruct the signal), the EBU source and licence, and a removal policy.
- The CI download step is removed. The gating and loudness-range stages are validated against the committed series, and the K-weighting front end against the synthesizable EBU cases in `test_program_loudness.py`, so CI needs no audio.
- The full signal chain against the original audio stays in `test_ebu_material_oracle.py`, which runs only where the set is present locally and skips otherwise.

https://claude.ai/code/session_01LnezkUn2wvXJ9LFKoYhdyd

## Summary by Sourcery

Document the EBU-derived loudness validation data and make the broadcast loudness tests in CI run without fetching external audio.

CI:
- Remove the CI step that caches and downloads the EBU loudness test set and rely solely on committed derived loudness series and synthesizable cases for CI validation.

Documentation:
- Add README documenting the EBU-derived per-block loudness series, their provenance, licensing context, and removal policy.

## Summary by CodeRabbit

* **Tests**
  * Updated CI loudness checks to rely on committed per-block measurement fixtures, so validation runs reliably offline without downloading or redistributing restricted audio assets.
* **Documentation**
  * Added documentation for the broadcast loudness derived fixture data, including sources, coverage, and how it’s used for validation.
  * Refreshed the testing guidance to explain when external reference material is required and why it’s not included in CI.
* **Chores**
  * Removed the workflow’s loudness test-set download/caching setup and related configuration.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -187,37 +187,13 @@ jobs:
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
         pip install -e .
-    # The authentic-programme loudness oracle (EBU Tech 3341 cases 7-8 and
-    # Tech 3342 cases 5-6) needs the official EBU loudness test set. Its
-    # licence only covers technical testing, so the audio is never committed:
-    # one matrix cell fetches it from the EBU (cached between runs) and the
-    # tests skip gracefully anywhere the set is absent.
-    - name: Cache the EBU loudness test set
-      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13'
-      id: ebu-cache
-      uses: actions/cache@v5
-      with:
-        path: ebu-loudness-test-set
-        key: ebu-loudness-test-set-v04-1
-    - name: Download the EBU loudness test set
-      if: >-
-        matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13' &&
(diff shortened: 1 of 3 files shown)


jmrplens:
Thanks. On the second point: the full-chain oracle's need for the EBU set (via `EBU_LOUDNESS_TEST_SET` or `plan/dsp-sources/`) is documented in the test's own skip message and in the new `tests/data/broadcast/README.md`, and I have now also added a note to `CONTRIBUTING.md` (Testing section) so developers running the suite locally find it there too.

On the first point I have kept the workflow comment inline: it explains, right where a reader would ask 'why is the EBU set not fetched here?', the licensing reason and what covers those cases instead; moving it behind a link would add indirection for little gain. It is seven lines and self-contained.

PR #250, opened 2026-07-20
feat: EN/ES language option for the aircraft/underwater/environmental plots

jmrplens:
## Summary

Every `.plot()` on the aircraft, underwater and environmental result objects now takes a `language` keyword. `"en"` stays the default and renders exactly as before; `"es"` renders localised axis labels, titles and legends and switches the decimal separator to a comma.

## What changed

- New private `phonometry._i18n` module with the shared helpers `check_language`, `format_number`, `decimal_comma` and `localize_axes`.
- The three `_plot` renderer modules (`aircraft.py`, `underwater.py`, `environmental.py`) gain a per-module `_STRINGS` translation table and a small `_t` lookup. Every fixed label, title, legend entry and annotation is now localised, keeping unit symbols and standard acronyms intact.
- The 31 delegating `.plot()` methods accept `*, language="en"`, validate it via `check_language` and forward it to the renderer.
- The `_i18n` import is done inside the functions so the plot modules keep no module-level cross-imports.
- Per-domain tests assert that `.plot(language="es")` yields Spanish labels and that an unknown language raises `ValueError`.

## English is unchanged

English strings are byte-for-byte identical to before. The two committed figures that actually exercise these renderers (`airport_contour`, `rotorcraft_terrain_screening`) were regenerated with zero diff to the committed SV/PNG assets.

## Checks

- `ruff check .`, `mypy src scripts` and the scoped test suite (`tests/aircraft tests/underwater tests/environmental tests/test_package_architecture.py`) all pass.
- API reference and `llms.txt` regeneration produced no drift.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Plot methods across aircraft, underwater, and environmental results now support English and Spanish localization.
  * Spanish plots include translated labels, titles, legends, annotations, and comma-based decimal formatting; English remains the default.
  * Unsupported language values are rejected with a clear error.

* **Documentation**
  * Updated API references and changelog entries to describe the new language option.

* **Tests**
  * Added coverage for English defaults, Spanish localization, localized legends, and invalid language handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   terminology and a comma decimal separator, while English stays exactly as
   before. A shared `phonometry._i18n` helper module backs the locale-aware
   number formatting and tick localisation.
+- The `.plot()` methods of the aircraft, underwater and environmental result
+  objects accept a `language` keyword (`"en"`, the default, or `"es"`). Spanish
+  renders localised axis labels, titles and legends with comma decimal
+  separators; English output is unchanged.
 - IEC 61260-1 filter class compliance can now be exported as a one-page PDF
   fiche. A new `filter_class_compliance(bank, *, num_points=..., edition=...)`
   runs the same verification as `verify_filter_class` and returns a frozen
--- a/site/src/content/docs/reference/api/aeroacoustics/aircraft-noise.md
+++ b/site/src/content/docs/reference/api/aeroacoustics/aircraft-noise.md
@@ -139,7 +139,12 @@ Effective Perceived Noise Level of an aircraft flyover (ICAO Annex 16).
 ### EPNLResult.plot()
 
 ```python
-EPNLResult.plot(ax: Axes | None = None, **kwargs: Any) -> Axes
+EPNLResult.plot(
+    ax: Axes | None = None,
+    *,
+    language: str = 'en',
+    **kwargs: Any,
+) -> Axes
 ```
 
 Plot the PNL and PNLT time histories with PNLTM and the 10 dB-down band.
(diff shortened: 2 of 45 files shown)


jmrplens:
Thanks for the review. Addressed in the latest push:

- **`localize_axes` (gemini + CodeRabbit).** Resolved by rebasing onto `main`, which now installs a `ScalarFormatter` subclass instead of the naive `FuncFormatter`. It preserves adaptive precision and offset notation, and it only runs when the current formatter is the automatic `ScalarFormatter`, so categorical axes (in matplotlib 3.11 `set_xticklabels` installs a `FuncFormatter`, not a `FixedFormatter`) and log axes are left untouched.
- **`ocean_ambient_noise` "Total" legend (CodeRabbit).** Localized via a new `total` string key, so `plot(language="es")` now shows the composite label in Spanish alongside `wind` / `thermal` / `shipping`.
- **Broader i18n test (CodeRabbit).** Added `test_ambient_noise_legend_localized`, a legend-heavy renderer that asserts every legend entry is localized (this is what would have caught the untranslated "Total").
- **`plot_transmission_loss` `:.3g` + `decimal_comma` (CodeRabbit, trivial).** Kept intentionally: the absorption coefficient spans orders of magnitude (roughly 1e-3 to 1e1 dB/km across the band), where 3 significant figures reads correctly and `decimal_comma` localizes the separator; `format_number`, with fixed decimals, would misrepresent the small-value end. In practice the value never reaches the scientific-notation range for physical seawater absorption.

PR #255, opened 2026-07-20
feat: EN/ES language option for the room/materials plots

jmrplens:
## Summary

Every `.plot()` in the room and materials domains now takes a `language`
keyword (`"en"` default, `"es"` for Spanish). With `language="es"` the axis
labels, titles and legends render in Spanish and linear-axis tick decimals use
a comma; the English output is unchanged. An unsupported code raises a clear
`ValueError`.

## Details

- New shared `phonometry._i18n` module holds the small locale helpers
  (`check_language`, `format_number`, `localize_axes`) used by the plot and
  report renderers. The plot layer imports it lazily inside functions, so it
  keeps no module-level dependency on it.
- `src/phonometry/_plot/room.py` and `src/phonometry/_plot/materials.py` gain a
  per-module string table and a small `_t` helper; each renderer threads
  `language` through, translating fixed strings and formatting embedded numbers
  with the locale separator.
- Each result's `.plot()` validates the language and forwards it to the
  renderer.
- Spanish acoustics terminology, for example "Coeficiente de absorción
  acústica", "Coeficiente de dispersión" and "Tiempo de reverberación".

## English output is unchanged

The default English path returns the verbatim original strings and formatting.
Rendering all 22 room/materials `.plot()` figures in English with the previous
code and with this change produces byte-for-byte identical SVGs.

## Tests

- New per-domain tests assert that every renderer yields a Spanish label under
  `language="es"` and rejects an unknown code with `ValueError`.
- The package-architecture test that forbids module-level imports in
  `_plot/*` still passes (helpers are imported lazily).
- `ruff check .` and `mypy src scripts` are clean; the API reference for the
  affected modules is regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added Spanish-language support for room and materials plots.
  * Spanish plots include translated labels, titles, legends, and comma-formatted decimal values.
  * English remains the default language.
  * Unsupported language codes now produce a clear error.

* **Documentation**
  * Updated plotting API references to include the new `language` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   `"es"`). Spanish renders localised axis labels, titles and legends and uses a
   comma decimal separator; English is unchanged. An unsupported language raises
   a clear `ValueError`.
+- The room and materials `.plot()` renderers accept a `language` keyword
+  (`"en"` default, `"es"` for Spanish). Passing `language="es"` renders the
+  axis labels, titles and legends in Spanish and switches linear-axis tick
+  decimals to a comma; the English output is unchanged. An unsupported code
+  raises a clear `ValueError`.
 - IEC 61260-1 filter class compliance can now be exported as a one-page PDF
   fiche. A new `filter_class_compliance(bank, *, num_points=..., edition=...)`
   runs the same verification as `verify_filter_class` and returns a frozen
--- a/site/src/content/docs/reference/api/materials/absorption-rating.md
+++ b/site/src/content/docs/reference/api/materials/absorption-rating.md
@@ -105,7 +105,12 @@ Weighted sound absorption rating (ISO 11654:1997).
 ### AbsorptionRatingResult.plot()
 
 ```python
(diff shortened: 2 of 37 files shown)


jmrplens:
Thanks for the review. Applied in the latest push (after rebasing onto `main`, which carries the improved `localize_axes`):

- **`_localize_band_axes` simplification (gemini).** In both `materials.py` and `room.py` the helper now takes just `(ax, language)` and only calls `localize_axes`, which reformats the automatic numeric y-axis and leaves the categorical band labels (a `FuncFormatter` on the linear position axis) untouched. Dropped the `labels` parameter, the fragile `get_xticklabels()` extraction and the label-restore workaround; call sites updated accordingly.
- **Comma decimal separator in embedded values (gemini).** Localized the numeric values baked into titles/labels via `decimal_comma`: `rating_label` (`0,65(H)`), the airflow evaluation velocity and `R_s` specific resistance, the porous-medium flow resistivity, and for consistency the mass-spring `m'` and the uncertainty coverage factor `k`. English output is unchanged.

620 room/materials tests pass, committed English figures are byte-identical, and ruff plus the package-architecture test are clean.

PR #260, opened 2026-07-20
Wire the EASA ANP fleet database into the ECAC Doc 29 airport-noise chain

jmrplens:
## Summary

I add `phonometry.aircraft.anp_fleet`, a loader that reads the EASA/EUROCONTROL Aircraft Noise and Performance (ANP) database tables and feeds real aircraft types into the existing ECAC Doc 29 airport-noise functions. Until now the Doc 29 chain (`npd_level`, `event_level`, `noise_contour`) accepted only synthetic NPD input; this lets me compute noise levels and contours from actual aircraft.

## Public API

- `load_anp_database(path=None)` returns an `AnpDatabase`. Without a path it loads the bundled full ANP database; pointed at a directory it reads any other ANP CSV export.
- `AnpDatabase`: `.aircraft(id)`, `.aircraft_ids`, `.npd_curves(id, operation, metric)`, `.profile(id, operation, stage_length)`, `.event_level(id, observer, operation, ...)`, `.noise_contour(id, operation, x=, y=, ...)`.
- `AnpAircraft`: aircraft metadata (description, engine type, engine count, weight class, Doc 29 `mounting`, NPD id) plus the same `npd_curves`/`profile`/`event_level`/`noise_contour` accessors.
- `AnpNpdCurves`: `LAmax`/`SEL` versus slant distance for each tabulated power setting, with `.level(power, distance)` and `.plot()`.
- `AnpProfile`: the default fixed-point trajectory as a ready `(N,5)` Doc 29 flight path `x, y, z, power, speed` (SI), with the derived takeoff `ground_roll` and landing `landing_roll` segment masks and `.plot()`.

`AnpDatabase`/`AnpAircraft` assemble the NPD curves and the profile and call the existing `event_level`/`noise_contour` unchanged, mapping the ANP lateral-directivity identifier to the Doc 29 engine mounting and the fixed-point points (feet, knots) to SI.

## How the database ships

The full EASA ANP database (archive version 2.3, 155 aircraft types) is bundled under `src/phonometry/aircraft/data/anp/` as package data, so the whole fleet is available out of the box. All ten v2.3 CSV tables are reproduced verbatim (upstream column layout and values unchanged, only the `ANP2.3_` filename prefix dropped) with provenance and licence in `data/anp/PROVENANCE.md`. The database is published by EASA/EUROCONTROL free of charge for noise/performance modelling and is redistributed on that non-commercial basis; it is not sold and carries no commercial benefit, and `load_anp_database(path=...)` can point at any other release.

The loader consumes the `Aircraft`, `NPD_data` and `Default_fixed_point_profiles` tables; the procedural-step, coefficient and spectral-class tables ship for completeness and future use.

## Scope

Only fixed-point trajectories are read as ready-to-use profiles. Modern types (A320, B738, B744, A388) ship procedural-step profiles that require the ICAO Doc 9911 / Doc 29 Vol 2 flight-mechanics performance model, which is out of scope here; requesting a profile for such a type raises a clear error, and NPD curves remain available for every aircraft.

## Oracle

Clean-room, against an independent published reference: the ANP database's own tabulated NPD values. Interpolating at a tabulated (power, distance) node recovers the published level exactly (delta 0). The test suite checks this round-trip exactness for the representative aircraft (both metrics, both operations, every distance node), which validates the parser and the log-linear/power interpolation, the main risk in wiring the real data in. A conformance row (`EASA ANP database round-trip`) pins Boeing 747-100 departure SEL at 28000 lb / 2000 ft = 98.8 dB.

The end-to-end contour still lacks an independent published numeric oracle (there is no worked ANP-to-contour footprint example in the sources at hand), so the contour path is covered only by a physical smoke test (finite, louder on-track than off-track). The per-term Doc 29 assembly it calls is already validated against the ECAC Vol 3 Part 1 reference workbook.

## Tests and gate

- `tests/aircraft/test_anp_fleet.py`: full-fleet load, round-trip exactness, metadata/mounting mapping, modern procedural-only aircraft (NPD yes, profile guard), profile units and ground-roll masks, event/contour smoke, external-directory load, error paths, EN/ES plot smoke.
- `ruff check .` clean, new module `mypy` clean, `pytest tests/aircraft tests/test_package_architecture.py` green, `docs/CONFORMANCE.md` + API docs + `llms` regenerated, `docs/api-reference.md` covers `__all__`, `check_figures.py` clean, data added to `MANIFEST.in`/package-data.

## Summary by CodeRabbit

* **New Features**
  * Added access to EASA ANP aircraft noise data, including aircraft lookup, NPD curves, and default flight profiles.
  * Added event-level noise calculations and noise contour generation using real aircraft data.
  * Added curve and profile plotting support.
  * Bundled a curated ANP dataset and support for loading full ANP exports.

* **Documentation**
  * Added API reference documentation and dataset provenance information.
  * Updated conformance results to include the new ANP database validation.


jmrplens:
Thanks for the review. I addressed all of the feedback in `ef95165`:

1. **BOM tolerance (`_read_tables`)** — the bundled and user-directory readers now use `encoding="utf-8-sig"`, so a leading BOM in an exported ANP CSV no longer mangles the first column. Covered by `test_utf8_bom_export_is_tolerated`.
2. **Unknown aircraft in `profile()`** — an unknown `aircraft_id` now raises the same "not in this ANP database (available: ...)" `KeyError` as `.aircraft()`, instead of the misleading "no fixed-point profile" message. Covered by `test_profile_unknown_aircraft_reports_missing_aircraft`.
3. **SEL/LAmax power pairing** — `_doc29_inputs` now validates that the SEL and LAmax curves share identical power settings (they always do in the ANP database; this guards a malformed user export) and raises a clear `ValueError` on mismatch before calling the Doc 29 chain. Covered by `test_mismatched_sel_lamax_powers_raise`.
4. **Profile-plot roll highlight off-by-one** — you were right, the shifted mask dropped the final point of a roll span. The highlight now marks both endpoints of every roll segment (`roll[:-1] |= seg; roll[1:] |= seg`), so the last runway point is included. Covered by `test_profile_plot_highlights_full_roll_span`.
5. **Ambiguous table match (`_pick`)** — when more than one file in an export directory matches a table keyword, it now raises an explicit "ambiguous ANP table" `ValueError` listing the matches instead of silently taking the first. Covered by `test_pick_ambiguous_table_raises`.
6. **Distance monotonicity (`_distances_m`)** — the parsed `L_<ft>ft` distance columns are now validated to be strictly increasing (as the dataclass docstring promises), raising `ValueError` otherwise.
7. **Read-only arrays** — the parsed NPD `powers`/`distances`/`levels` and the profile `path` are now exposed as read-only views (`.flags.writeable = False`, no copies), which matters because `distances` and the NPD/path arrays are shared across results; the docstrings note it. The Doc 29 chain reads (never mutates) these, so nothing downstream breaks. Covered by `test_parsed_arrays_are_read_only`.

Gate re-run green with `PYTHONPATH` set to the worktree src: `ruff check .`, `mypy` (module clean), `bandit` (clean), `pytest tests/aircraft tests/test_package_architecture.py` = 272 passed / 10 skipped; API docs and llms regenerated.

PR #261, opened 2026-07-20
Objective impulsive-sound prominence and LAeq adjustment (ISO/PAS 1996-3:2022)

jmrplens:
## Summary

I add `environmental.impulsive_sound`, the objective method of ISO/PAS 1996-3:2022 for measuring the prominence of impulsive sounds and deriving the adjustment `KI` to `LAeq`. It complements the existing `impulse_prominence` (NT ACOU 112), which takes the onset rate and level difference as inputs: ISO/PAS 1996-3 keeps the same prominence and adjustment formulae but adds the objective measurement chain that reads those quantities directly from a calibrated recording.

## Method

`impulsive_sound_adjustment(signal, fs)`:

1. A-weights the signal (IEC 61672-1) and applies time weighting F (tau = 125 ms) to build the level history `LpAF`, sampled every 10-25 ms (Clause 4).
2. Detects onsets: a starting point where the gradient first exceeds 10 dB/s, an end point where it drops below, merging events less than 50 ms apart (Clause 3.3, procedures a-d of Clause 4).
3. Measures each onset's level difference `LD = Le - Ls` (3.4) and its least-squares onset rate `OR` (3.5), with an optional upper-half variant for pass-bys (3.5, Note 1).
4. Evaluates the prominence `P = 3 lg(OR) + 2 lg(LD)` (Clause 5) and returns the governing adjustment `KI = 1.8 (P - 5)` dB for `P > 5`, else 0 dB (Clause 6).
5. Categorises the source (Clause 7): not impulsive (`KI = 0`), regular impulsive (`0 < KI <= 5`) or highly impulsive (`KI > 5`), and reports the adjusted `LAeq`.

**Adjustment range.** The scope states adjustments typically fall between 0,0 dB and 9,0 dB. The formula of Clause 6 is not itself capped (the standard notes `P` is designed to peak around 15), so the implementation applies it faithfully without an artificial ceiling and documents this.

## API

- `impulsive_sound_adjustment(signal, fs, *, dt=0.02, reference_pressure=2e-5, calibration_offset=0.0, onset_rate_method="least_squares", laeq=None) -> ImpulsiveSoundResult`
- `sound_pressure_level_history(signal, fs, ...) -> (times, levels)`
- `detect_onsets(levels, dt, *, onset_rate_method=...) -> tuple[ImpulseOnset, ...]`
- `ImpulsiveSoundResult` (frozen; `.plot()`, `.governing_onset`), `ImpulseOnset`, `ImpulsiveSoundWarning`

All re-exported from the flat `phonometry` namespace and the `environmental` subpackage.

## Oracle

ISO/PAS 1996-3 gives no worked numeric example (confirmed by reading the full document, including the annexes), so I verify against an oracle derived by hand from the standard's own definitions, independent of the implementation: a straight `LpAF` ramp of 30 dB over 0.30 s has, by construction, level difference `LD = 30 dB` and, since the least-squares slope of a straight ramp equals its slope, onset rate `OR = 100 dB/s`. Therefore `P = 3 lg(100) + 2 lg(30) = 8.9542426` and `KI = 1.8 (P - 5) = 7.1176366 dB`. Two conformance checks and the test suite assert these hand values. Boundary tests cover a steady interval (no onset, 0 dB, not impulsive) and a very sudden ramp (highly impulsive), plus the 50 ms onset merging, the pass-by variant, and the full signal chain on synthesized signals (a steady tone yields 0 dB; an impulsive burst is detected and highly impulsive; the adjustment is invariant to a calibration offset, per Clause 8).

## Verification

- `ruff check .` clean; `mypy` clean on the new module and edited scripts (the pre-existing numpy-typing errors on `main` are unrelated to this change and come from a local numpy version older than CI's).
- `pytest tests/environmental` (241) and `tests/test_package_architecture.py` (22) pass, plus 21 new tests.
- `scripts/check_api_reference.py`, `scripts/check_figures.py` (zero diff) and `site/scripts/check-i18n-parity.mjs` all pass; API docs and llms regenerated.
- Docs: EN and ES guide sections, the API-reference rows, the bibliography note and a CHANGELOG entry.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added objective impulsive-sound assessment aligned with ISO/PAS 1996-3:2022.
  * Detects sound onsets, calculates prominence and adjustment, categorizes impulsiveness, and provides adjusted LAeq results.
  * Added optional plotting of level history, detected onsets, and governing measurements.

* **Documentation**
  * Added API references, usage guides, examples, and standards documentation in English and Spanish.
  * Added conformance coverage and release notes for the new functionality.


jmrplens:
Thanks both.

- **gemini (onset trend line):** fixed. The dashed line now anchors the fitted slope at the onset-span midpoint (`l_mid + onset_rate*(t - t_mid)`) rather than forcing it through `(time_start, level_start)`. Since the least-squares regression passes through the sample centroid (~the midpoint), the drawn trend no longer overshoots the marked endpoints, including the `"upper_half"` variant.
- **Sonar S5778:** hoisted the signal/result construction out of the four flagged `pytest.raises` blocks so only the call under test can raise.
- **CodeRabbit (`_equivalent_level` re-A-weights):** declining this one. Deduplicating the A-weighting would require either changing the public `sound_pressure_level_history` signature to accept a pre-weighted signal or duplicating its internal weighting/sampling logic in the caller; the second A-weighting is a single linear filter pass and the coupling/API cost is not worth it for the targeted interval sizes. Left as is for clarity.

PR #265, opened 2026-07-20
Loudspeaker rated-characteristics report (IEC 60268-5, graphs to IEC 60263)

jmrplens:
## Summary

I add a loudspeaker measurement/rating report for the rated characteristics IEC 60268-5:2003+A1:2007 defines around a measured on-axis response, with the characteristic frequency and polar graphs drawn to the IEC 60263:1982 scale conventions.

`electroacoustics.loudspeaker_characteristics` returns a `LoudspeakerCharacteristics` result that gathers the rated impedance, the rated frequency range, the rated powers, the resonance frequency, the impedance modulus curve, the total-harmonic-distortion curve and the directional/polar response. Two of the rated characteristics are computed from the on-axis response rather than merely repeated:

- **Characteristic sensitivity level** (clauses 20.3/20.4): the on-axis level averaged over a stated band, referred to 1 W into the rated impedance at 1 m, with `Up = sqrt(R)` (the 2,83 V into 8 ohm convention as the default drive).
- **Effective frequency range** (clause 21.2): the band over which the response stays within 10 dB of the level averaged over the one-octave band in the region of maximum sensitivity, neglecting troughs narrower than 1/9 octave.

The distortion panel reuses a `SweptSineDistortionResult` and the polar panel a `RadiatingPistonResult` directivity.

`LoudspeakerCharacteristics.report(path, *, metadata=None, engine="reportlab", verbose=False, language="en")` renders the one-page rated-characteristics fiche on the shared report engine: the rated-characteristics table beside the on-axis response with its tolerance band and effective-range markers, plus the impedance (with the 80 %-of-rated line), THD and polar-directivity panels. The response and polar graphs follow IEC 60263 (one frequency decade equal to 25 dB on the ordinate; the polar diagram on a 25 dB reference-circle radius). The fiche is localised in English and Spanish.

## Oracle and conformance

- Tests check the two computed characteristics against the standard's own definitions as clean-room oracles: a flat `L0` response driven at `sqrt(R)` volts at 1 m returns `L0` exactly, a doubled drive voltage returns `L0 - 6,02` dB, and a response crossing the `reference - 10 dB` threshold at chosen frequencies returns exactly those frequencies (with a narrow trough neglected).
- Two conformance rows cover the characteristic sensitivity level and the effective frequency range.

## Deliverables

- `src/phonometry/electroacoustics/loudspeaker.py` (model) and `src/phonometry/_report/iec60268_5.py` (renderer), additive `_report/_i18n.py` strings.
[...]

PR #267, opened 2026-07-20
Microphone rated-characteristics report (IEC 60268-4, graphs to IEC 60263)

jmrplens:
Add a microphone measurement/rating report for the rated characteristics IEC 60268-4:2014 defines around a measured free-field frequency response, the companion of the loudspeaker fiche (#265), with the characteristic frequency and polar graphs drawn to the IEC 60263:1982 scale conventions.

`electroacoustics.microphone_characteristics` builds a `MicrophoneCharacteristics` result that gathers the rated free-field sensitivity, the response tolerance, the rated and minimum permitted load impedances (10.2/10.3), the overload sound pressure level at a stated THD limit (14.2/15.2), the inherent-noise data (17), the directional pattern (13.1) and the rated power supply (9.1). Four of the characteristics are computed from the standard's own definitions rather than merely repeated:

- **Sensitivity level** (11.1/11.3): `L_M = 20 lg(M / 1 V/Pa)` of the rated free-field sensitivity at the 1 kHz reference frequency.
- **Effective frequency range** (12.2): the interpolated crossings of the response, normalized to 0 dB at the reference frequency, with the +/- tolerance limits.
- **Directivity index** (13.2.2): `D = 20 lg(M0 / Mdiff)` with the diffuse-field sensitivity from the 11.2.2 a) integral over a rotationally symmetric pattern.
- **Equivalent sound pressure level due to inherent noise** (17.2 d/e): the weighted noise voltage over the rated sensitivity, re 20 uPa, with the signal-to-noise ratio re 1 Pa and the diffuse-field sensitivity level (11.2.2.1) derived from it.

`MicrophoneCharacteristics.report()` renders the one-page rated-characteristics fiche: the rated-characteristics table (sensitivity in mV/Pa and its level re 1 V/Pa, impedances, equivalent noise level, SNR, maximum SPL, directivity index, phantom powering) beside the free-field response with its tolerance band, reference-frequency and effective-range markers, plus the directional-pattern, inherent-noise-spectrum and distortion-against-level panels. The response and polar graphs follow IEC 60263: one frequency decade equal to 25 dB on the ordinate, and the polar diagram on a 25 dB reference-circle radius, sharing the graph helpers of the loudspeaker renderer. The fiche is localised in English and Spanish, and a `requirement` in the metadata is checked as a maximum permitted equivalent noise level.

Verify the four computed characteristics against the standard's definitions as clean-room oracles, in the tests and as four conformance rows:

- 12.5 mV/Pa returns `20 lg 0.0125 = -38.0618` dB re 1 V/Pa (hand-computed).
- A response crossing the +/-3 dB limits at chosen frequencies returns exactly those frequencies as the effective range, on the lower and the upper limit.
- The ideal cardioid returns `D = 10 lg 3 = 4.7712` dB from the diffuse-field integral.
- 2.5 uV over 12.5 mV/Pa is 200 uPa, i.e. exactly 20.0 dB SPL, and a distortion curve reaching the 0.5 % limit at 130 dB places the overload SPL there exactly.

Add a committed example fiche, the English and Spanish guide sections with a report preview, the IEC 60268-4 bibliography entries, and the regenerated API docs and conformance report.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added IEC 60268-4:2014 microphone characterization from measured response data.
  * Calculates sensitivity, effective frequency range, directivity, noise, SNR, overload level, impedance, and power requirements.
  * Generates one-page rated-characteristics reports in English or Spanish, including optional response, polar, noise, and distortion plots.

* **Documentation**
  * Added API reference, guides, bibliography entries, examples, and updated conformance reporting for microphone characterization.


jmrplens:
Addressed the review findings in 982d81b:

- **Directivity integral vs 0..360 angles (gemini, CodeRabbit)**: fixed. The 11.2.2 a) integral runs over the polar angle 0..180, so `_directivity_index_from_polar` now folds angles beyond 180 onto `360 - angle` (rotational symmetry about the reference axis) before integrating; duplicated folded angles form zero-width trapezoid segments and contribute nothing extra. A full-circle cardioid (0..360) now returns the same `10 lg 3 = 4.7712` dB as the half-circle input (test added), and the computation is gated on the *folded* angular span, so a front-only arc supplied as 270..360 no longer produces a bogus index (test added).
- **Empty `polar` arrays**: now rejected with a clear `ValueError` ("'polar' needs at least two angle points") before any indexing (test added).
- **Non-finite stated `directivity_index_db`**: the finiteness check now runs before the `polar is None` branch, so it applies whether or not a pattern is supplied (test added for both branches).
- **Empty `distortion` / `noise_spectrum` and non-positive `frequencies`**: these were already rejected by the shared curve validator (`_as_curve` requires two points and positive, finite frequencies), so no crash was reachable through the public builder; I added a defensive size guard in the overload-SPL reader anyway and locked the three rejections in with explicit tests.
- **Sonar S107 (too many parameters)**: declined, same rationale as the loudspeaker sibling (#265): the builder's keyword-only parameters map one-to-one onto the rated characteristics IEC 60268-4 defines (impedances 10.2/10.3, sensitivity 11, response tolerance 12, directional data 13, distortion 14/15, inherent noise 17, powering 9.1), and splitting them into an options object would only move the same surface one level down while breaking the uniform `*_characteristics(...)` builder convention.

The rendered example fiche is unchanged: its pattern only spans 0..179, for which the folding is an identity, and the computed DI is bit-identical.

PR #272, opened 2026-07-20
Fix the ground-impedance time convention (restores the soft-ground dip) and add the GFPE surface-wave term

jmrplens:
## The defect

The porous models of `phonometry.materials` (`delany_bazley`, `miki`) work in the `e^{+jwt}` time convention, in which a passive medium has `Im(Z) < 0`. The Weyl-Van der Pol and GFPE formulas of the environmental domain (`F(w)` through `wofz`, `exp(+1j k (R2 - R1))`, `R(kz)`) follow Salomons' `e^{-iwt}` convention, in which a passive ground has `Im(Z) > 0`. `_normalized_ground_impedance` fed the materials impedance across that boundary unconjugated, so the phase of the spherical-wave reflection coefficient `Q` was wrong and the soft-ground interference dip nearly vanished. The docstrings compounded it by telling users to pass `Im(Z) < 0` "in the `e^{-iwt}` convention".

Reproduction in the Salomons Fig. D.3 geometry (grassland `sigma = 200 kPa s/m2`, `hs = hr = 2 m`, `r = 100 m`): `ground_effect` gave a **-0.55 dB** minimum at 360 Hz where the correct dip is **-12.72 dB near 395 Hz**. The independent Delany-Bazley + Chien-Soroka implementation in `aircraft/rotorcraft_noise.py` (which already uses the `Im(Z) > 0` pairing) agrees with the corrected result within 0.3 dB.

## The fix

- `_normalized_ground_impedance` conjugates anything obtained from the porous models (the `flow_resistivity=` path and any `PorousMediumResult`), converting `e^{+jwt}` to `e^{-iwt}` before it enters the formulas. This feeds `ground_effect`, `spherical_reflection_coefficient` consumers, the coherent ground model of `barrier_insertion_loss` and the PE ground condition.
- One documented convention for user-supplied impedance values: `e^{-iwt}`, `Im(Z) > 0` for a passive ground, matching Salomons and the rotorcraft module. The wrong docstrings in `ground_barriers.py` and the `atmospheric_refraction.py` module header are corrected, and the guides (EN/ES), `docs/` mirrors and regenerated API reference now state the convention and the internal conjugation explicitly.
- The GFPE range step now includes the **surface-wave term** (the residue of the ground-reflection pole at `kz = -k0/Z`, third term of Salomons Eq. (H.49)). The residue contour encloses that pole exactly when `Im(Z) > 0`, so the term is applied for a passive finite-impedance ground and is absent for a hard or non-passive one. With the corrected convention but without this term, the homogeneous-limit PE drifted up to about 1 dB from the Weyl-Van der Pol oracle; with it the agreement is back to a few tenths of a dB on the default grid.

## Tests and oracles

- New oracle test and conformance row pinning the Fig. D.3 dip: minimum `dL = -12.7 +/- 0.3 dB` within 380-410 Hz (computed: **-12.72 dB @ 395 Hz**).
- Convention-consistency tests: `impedance=delany_bazley(...)` equals the `flow_resistivity=` path, and a hand-conjugated `e^{-iwt}` array equals both.
- PE soft-vs-hard ground test: at the dip frequency and low heights the finite-impedance PE level sits more than 8 dB below the hard-ground level at 100 m.
- Tests that pinned the inverted-convention behaviour are updated with their physical anchors stated in comments (`|Q|` may now slightly exceed 1 at low frequency near grazing: the surface-wave contribution, Salomons Sec. 3.4).
- Conformance: 368/368 checks pass; the GFPE-vs-ground-effect row moves from -13.9 dB to -16.4 dB (both sides recomputed, delta 0.035 dB).

## Figures

The physics change moves two committed figure families (all four language/theme variants each): `ground_effect_spherical*.svg` (the four ground types now show their true dips, e.g. grassland -12 dB near 540 Hz for the 1 m / 1.5 m / 50 m geometry) and `atmospheric_refraction*.png` (the GFPE panel over grassland). Alt text and prose quoting the old curve minima are updated in EN and ES. `scripts/check_figures.py` passes against the committed set.

## Gates

`ruff check .`, `mypy src scripts`, `bandit -r src` (0 issues), full `pytest` (4314 passed, 64 skipped), `scripts/check_figures.py`, conformance and API-reference/llms regeneration, `site/scripts/check-i18n-parity.mjs` (90/90).

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected ground-reflection and interference calculations for porous-material impedance values.
  - Improved atmospheric propagation over finite-impedance ground by including surface-wave effects.
  - Restored expected soft-ground interference dips and reduced propagation-level drift.

- **Documentation**
  - Clarified passive-ground impedance conventions and handling of porous-model inputs.
  - Updated validation results, examples, figures, and conformance coverage.


jmrplens:
Merging. On the CodeRabbit note (impedance validation flow): declining as designed — the `flow_resistivity` path builds the impedance from the library's own porous models, which are finite and non-zero by construction; the `isfinite`/`|Z|>0` guards exist for the user-supplied array path, which is where malformed input can enter. Keeping the early return avoids re-validating internally generated values on every call.

PR #281, opened 2026-07-21
Correct fiche verdict rules and mandatory report items across the .report() layer

jmrplens:
The `.report()` fiches carried a handful of places where the printed sheet did not say exactly what the underlying standard says, plus one wrong compliance rule. This fixes them and pins the displayed numbers in tests.

## EBU R 128 tolerance (the important one)

The programme-loudness fiche passed the verdict on a blanket +/-0.5 LU about the target. That was the June 2014 (V3) rule; R 128 dropped it in the August 2020 revision and R 128-2023 keeps only two tolerances: +/-0.2 LU for measurement errors in loudness workflows such as Quality Control (item i) and +/-1.0 LU where the Target Level is not achievable practically, e.g. live programmes (item h). A non-live programme at -23.5 LUFS rendered PASS under a rule that no longer exists.

`ProgramLoudnessResult.report()` now takes `tolerance="qc"` (default, +/-0.2 LU per item i) or `tolerance="live"` (+/-1.0 LU per item h), prints the applied rule and its R 128 item on the fiche, and evaluates the verdict on the loudness rounded to the displayed 0.1 LU (EBU Tech 3341 requires a display precision of at most one decimal place), so the printed numbers can never contradict the verdict at the boundary. The committed example signal is trimmed 0.4 dB so it sits on the -23.0 LUFS target and passes the QC rule it now demonstrates.

## Spanish fiches now embed Spanish charts

`language="es"` localised every fixed string and number but called the result's `plot()` without forwarding the language, so the embedded chart kept English axis labels and legends. The fiche language is now forwarded to `plot_fn` (every result plot accepts `language`), verified visually on the rendered ES fiches.

## Per-standard corrections

- **ISO 532-1 (clause 7)**: the fiche now states the method used (stationary sounds, clause 5, or time-varying sounds, clause 6) and the sound field (free F / diffuse D; `ZwickerLoudness` gains a `field` attribute set by the constructors), renders the mandatory loudness-versus-time function N(t) for time-varying sounds, and labels the reported maximum "Maximum loudness Nmax" as clause 7 f) names it.
- **ICAO Annex 16 EPNL**: `effective_perceived_noise_level()` gains `procedure="aeroplane"|"helicopter"`, plumbing the App. 2 4.3.1 Step 1 tone-correction start band (50 Hz for helicopters and tilt-rotors) through the public chain; previously a helicopter flyover silently used the aeroplane procedure and could miss rotor tones in the 50-80 Hz bands. The verdict now determines the EPNL once to one decimal place, as the Annex states it, before deriving the level, margin and status (101.04 vs a 101 limit is 101.0, margin +0.0, PASS, instead of a FAIL contradicting its own printed numbers). The reference-conditions strip now cites Part II 3.6.1.5 with the full set (25 C, i.e. ISA + 10 C, 70 % RH, 1013.25 hPa, zero wind).
- **ISO 717**: an octave-band rating was captioned "One-third-octave"; the caption now declares the band set actually rated (clause 4.4 requires stating it) and the octave table drops the third-octave triplet rules. `report()` gains `symbol=` ("R'w", "Dn,w", "DnT,w", "L'n,w", "L'nT,w", ...) so a field measurement is not mislabelled with the laboratory Rw / Ln,w descriptor; the symbol heads the boxed result, the value-table column and the verdict. Adaptation terms print a sign only when negative, the style of the standard's own examples ("41 (0; -5) dB").
- **IEC 61260-1**: the multirate verifier cannot exercise the stop-band mask beyond each band's processing Nyquist, so `verify_filter_class()` / `FilterComplianceResult` now report `range_limited` (plus per-band `checked_to_omega`) and the fiche qualifies its COMPLIES statement instead of asserting unqualified full-mask conformance. Bands are labelled with the nominal mid-band frequencies both editions identify filters by (125 Hz, not 125.89.. Hz), the 1995 fiche cites that edition's own relative-attenuation definition (3.13 Note) instead of a 2014-only formula number, and a `required_class` that does not exist in the verified edition (class 0 vs 2014) is rejected instead of rendering a meaningless FAIL.
- **ISO 11654**: the boxed rating writes the shape indicator in the clause 5.3 style, "0.60(M)" with no space, and prints the 5.3 NOTE recommendation (use alpha_w together with the complete curve) when an indicator applies.

## Cross-cutting polish

- Fiche header grids reprint client-supplied metadata verbatim (an area of 1.23 m2 was display-rounded to "1.2").
- Values that round to zero at the displayed precision never print "-0.0".
- ES fiches localise the remaining English fragments ("Table 1" -> "Tabla 1", "1/1-octave" -> "1/1 de octava").

## Tests

Every fiche now has at least one pypdf content test that extracts the rendered text and asserts the exact displayed values (tolerance and clause on the EBU fiche, "30 (-2; -3) dB" on ISO 717, "0.60(M)" on ISO 11654, nominal band labels and the range qualification on IEC 61260, method/field/Nmax/N(t) on ISO 532-1, the one-decimal EPNL boundary), plus boundary tests for the display-rounded verdicts and the helicopter start band (a 63 Hz tone yields C > 0 only under the helicopter procedure).

Committed example fiches, the API reference and llms files are regenerated; the programme-loudness guide and CHANGELOG are updated (EN + ES).

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added selectable **QC** and **Live** tolerances for EBU R 128 programme-loudness reports.
  * Added **helicopter** tone-correction support for EPNL effective perceived noise level.
  * Added **ISO 717** report symbol customization for rendered single-number labels.
  * Added sound-field capture for **ISO 532-1** Zwicker loudness results and expanded ISO 717/532 reporting details.
  * Added filter-class **range-limited qualification** to compliance outputs/reports.
* **Bug Fixes**
  * Improved verdict logic and numeric rounding consistency across fiches and localization previews.
* **Documentation**
  * Updated API and guide texts (including Spanish) for the new options, tolerance rules, and reporting contents.


jmrplens:
Addressed the SonarCloud gate and the review comments in the latest push:

- **S1244 (both):** the signed-zero normalisation in `format_number` now inspects the formatted text (no float comparison), and `display_round` compares its non-negative magnitude with `<=`.
- **S3776 (both):** extracted `_verify_band` from `verify_filter_class` and `_basis_line` from `render_iso532_report`; behaviour identical, tests untouched and passing.
- **S5778 (x4):** hoisted the setup calls out of the flagged `pytest.raises` blocks.
- **CodeRabbit, broadcast.py:** the live tolerance now prints through `format_number(..., decimals=1)`, so "1.0 LU" keeps its decimal in both languages.
- **CodeRabbit, iso532.py:** the metrics-table loudness row now applies the same `display_round` as the boxed statement, so the two cannot disagree at a rounding tie.
- **CodeRabbit, test_absorption_rating.py:** the parametrised case already documents the binary-representation nuance in detail (0.325*100 sits just above 32.5 in binary; exact hundredth-halves cannot arise from 2-decimal ISO 354 inputs), so the test is left as is.

PR #287, opened 2026-07-21
Migrate committed raster doc assets from PNG to lossless WebP

jmrplens:
The committed raster documentation assets move from optimized PNG to lossless WebP. The pixels are identical (verified with a per-channel decode comparison against every old PNG: max difference 0.0), only the container changes. SVG figures, animations, posters and the og-image are untouched.

## What changed

- **Report-fiche previews** (`.github/reports/*.webp`, 13 files): `scripts/generate_reports.py` now saves the pypdfium2 raster directly as lossless WebP (`lossless=True, quality=100, method=6`) instead of PNG + oxipng. `preview_path_for` pairs `.pdf` with `.webp`.
- **Raster figures** (`.github/images/*.webp`, 11 figures x 4 light/dark/EN/ES variants = 44 files): `save_figure` in `scripts/generate_graphs.py` renders the matplotlib canvas to an in-memory PNG (metadata stripped, as before) and re-encodes those exact pixels as lossless WebP. The `_PNG_FIGURES` set is now `_RASTER_FIGURES`.
- `scripts/check_figures.py` applies the same pixel-tolerance compare to `.webp` that it used for `.png` (Pillow decodes both), so the figures CI gate still catches real drift while tolerating cross-CPU 1 ULP noise.
- All embeds updated: site guides (EN + es/), `docs/*.md` picture tags, the `ReportPreview` component (it derives the image URL from the fiche name, so call sites needed no change) and `ThemeImage` call sites. Zero references to the deleted PNGs remain.
- The now-unused `_optimize_png` helpers and the `pyoxipng` pin (requirements-dev, requirements-figures) are removed.
- `tests/test_generate_reports.py` asserts a valid 1000 px wide WebP preview per example instead of the PNG magic check.

## Measured sizes

| Asset set | Before (PNG) | After (WebP) | Saved |
|---|---|---|---|
| 13 report previews | 1,672,980 B | 820,838 B | 852,142 B (50.9 %) |
| 44 raster figures | 26,160,244 B | 17,075,950 B | 9,084,294 B (34.7 %) |
| Total | 27,833,224 B | 17,896,788 B | 9,936,436 B (35.7 %) |

## Determinism

Pillow's lossless WebP encode at `method=6` is byte-stable: three independent full runs of the report generator and two full runs of the raster figures produced byte-identical files, so `make graphs` / `make reports` stay reproducible and diff-clean. The regenerated SVGs also came out byte-identical to the committed ones (untouched by this PR).

## Checks

- ruff, mypy (src + scripts), bandit: clean
- `pytest tests/test_generate_reports.py tests/test_conformance_report.py` (397) plus `-k report` across the suite (534): all pass
- `python scripts/check_figures.py`: all 892 committed figures match
- `node site/scripts/check-i18n-parity.mjs`: 96/96 pages
- Full site build + html-validate on the 32 touched guide pages (EN + ES): clean; built pages reference the `.webp` assets

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jmrplens:
Applied all three suggestions: the in-memory render buffer is now managed by a with block, and the preview tests read _PREVIEW_WIDTH_PX from the loaded generator module instead of duplicating the constant. Determinism re-verified (check_figures passes on the committed WebP set).

PR #377, opened 2026-07-27
Set up Discussions: category forms, issue contact links and a code of conduct

jmrplens:
Discussions is enabled but still on the default categories, with nothing routing people to it. This adds the parts that live in the repository.

## Category forms

`.github/DISCUSSION_TEMPLATE/`, one file per category slug:

- `q-a.yml`: what you are trying to compute, what you tried, version and environment. Points clause-level questions at the standards category and reproducible defects at Issues.
- `standards-conformance.yml`: designation with edition and year, clause or table, the expected value **and where it comes from**, and the snippet that produces the value obtained. It also asks people not to paste substantial verbatim text from copyrighted standards, and links `docs/CONFORMANCE.md` and `docs/ERRATA.md`.
- `documentation-learning.yml`: which page, what was confusing, and optionally the reader's background so the level of an explanation can be calibrated. English and Spanish both welcome.
- `ideas.yml`: the problem, the document that defines the method, and the reference data it could be validated against. That last field is usually what decides whether a proposal is actionable.
- `show-and-tell.yml`: what was built, which modules were used, and the Zenodo DOI for citation.

Forms are matched to categories by file name, so they only take effect once the categories exist.

## Routing

`.github/ISSUE_TEMPLATE/config.yml` sends questions, standards queries, proposals and documentation feedback to their categories before a blank issue is opened. Blank issues stay enabled for real bugs.

## Community files

- `CODE_OF_CONDUCT.md`, Contributor Covenant 2.1. GitHub surfaces it in the Discussions sidebar, and the repository had none.
- `CONTRIBUTING.md` gains a "Where to ask" table, plus the two conventions that matter here: cite clauses instead of pasting them, and name the validation data when proposing a method.

A `conformance` label was created for the standards category.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/377?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by Sourcery

Set up GitHub Discussions workflows, templates, and community guidelines to route questions and proposals appropriately and clarify expected conduct.

Enhancements:
[...]

PR #379, opened 2026-07-27
Add issue forms and label pull requests from the paths they touch

jmrplens:
Follow-up to #377. Discussions now has its forms and its routing; this gives Issues the same, and makes labels something that actually gets applied.

## Issue forms

`.github/ISSUE_TEMPLATE/`, scoped so the routing in CONTRIBUTING holds rather than being restated and contradicted:

- **Bug report**: a crash, a result that contradicts the library's own documentation, or a broken installation. A required checkbox confirms it is not a disagreement with a standard, which is the case that would otherwise land here by default.
- **Conformance defect**: a discrepancy already established, normally in a Standards & conformance discussion, with a field for that link. Asks for the required value and the computed value together with the input conditions and the tolerance the standard allows, and asks whether the defect looks like it is in the library or in the published standard, which is the fork between a fix and an `docs/ERRATA.md` entry.
- **Implement a standard or method**: a work item, not a proposal. Scope has to name the excluded clauses too, since that is what keeps the conformance claim honest, and the reference data field is required because an item without an oracle stays blocked regardless.
- **Documentation defect**: concrete defects only, with unclear explanations pointed at Documentation & learning.

## Labels applied from paths

None of the last 15 pull requests carried a label, so a scheme that needs manual application would not have changed anything. `.github/labeler.yml` derives them from the changed paths instead:

- **Kind of change**: `documentation`, `site`, `i18n`, `figures`, `reports`, `conformance`, `errata`, `validation`, `api`, `packaging`, `ci`.
- **Acoustic domain**: one `area:` label per package under `src/phonometry/`. `tests/` mirrors those names, so an implementation and its tests land on the same label.

`sync-labels` is off, so labels added by hand are never removed. The workflow runs on `pull_request_target` because the default token is read-only for fork pull requests under `pull_request`; the action only reads the changed file list and never checks out or runs the pull request's code.

The 26 new labels are already created in the repository. Every label referenced by a form or by the labeler was checked to exist, and every glob was checked to match at least one real path.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/379?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by Sourcery

Introduce structured issue templates and automatic path-based labelling for pull requests.

CI:
- Add a GitHub Actions workflow to apply labels to pull requests based on changed file paths.
- Configure path-based label mapping for documentation, CI, packaging, and acoustic domain modules.

Documentation:
- Document available issue forms and automatic labelling in CONTRIBUTING.


jmrplens:
Reviewed independently and pushed the fixes in af98e15b.

Three coverage holes in the labeler, all confirmed against the tree:

- `.github/reports/**` (128 files): a pull request that only regenerated the fiches carried no `reports` label.
- `.github/scripts/**` and `.github/dependabot.yml` missing from `ci`.
- The one that mattered: `src/phonometry/_internal/**`, `_compat.py`, `_version.py`, `_i18n.py` and the 28 cross-cutting suites at the root of `tests/` matched nothing. That is code every domain package rests on, going out with no area label. Now covered by a new `area: core` label. After the fix the only uncovered production path is `py.typed`, an empty marker.

Two placeholders named things that do not exist:

- `phonometry.levels.equivalent_level` is not a real symbol; the public name is `phonometry.leq`. It was wrong in the discussion form on main as well, so both are corrected.
- The new-standard form used ISO 3382-2 and `reverberation_time` as its example, and both are already implemented. Anyone checking whether their proposal was a duplicate got the opposite signal. Replaced with ISO 3746, which has no conformance row.

Also: when code and documentation disagree the reporter cannot tell which is wrong, so that case now defaults to a bug report and the documentation form is reserved for when the text is clearly the mistaken part; and the bug form no longer demands a Python snippet and an installed version from someone whose install failed.

Outside the diff, on the label set: `github-actions`, `python` and `javascript` are referenced by `.github/dependabot.yml` but never existed, so Dependabot has been dropping them silently and the per-ecosystem routing never worked. Created. `performance` had nothing applying it and now has a rule. `ci` was `#ededed`, indistinguishable from an uncoloured label. `improvement` duplicated `enhancement` and was removed after moving its single use, and the `Review effort 2/5` residue of a retired bot was removed too.

Not taken: SHA-pinning `actions/labeler@v7`. Stricter, but all seven workflows here use major tags with Dependabot keeping them current, and breaking that in one file is not worth it.

PR #395, opened 2026-07-27
GPU runner scaffolding for the FDTD animation fields

jmrplens:
Add a backend-injectable replica of the FDTD2D leapfrog engine
(scripts/fdtd_gpu.py) that runs the same stepping code on NumPy or CuPy
for the subset the animations use: per-cell c/rho maps, rigid walls,
sponge layers, per-side real-impedance edges, scalar or per-cell
damping, rasterised obstacles and the one-way plane-wave initial
condition. With xp=numpy the pressure history is bit-identical to the
library engine; tests/test_fdtd_gpu_parity.py locks parity per feature
and per travel direction (including heterogeneous c maps, right/bottom
impedance edges, a bare-string sponge side and the explicit empty
sponge_sides tuple), and the CuPy cases run automatically when a CUDA
device is present.

scripts/fdtd_gpu_remote.py packs a job (maps, parameters and a frame
sampling schedule) into a single npz, ships it with the engine to a
CUDA host over SSH and runs it there with Docker (--gpus all) through
scripts/job_runner.py, fetching the sampled frames back and cleaning
the per-job directory. Jobs can request device-side frame subsampling
(sample_stride) and float32 casting (sample_dtype) before the transfer,
plus an optional init_scale_x window that tapers the initial front
laterally (p and vy column-wise). build_job validates its inputs like
the library (strict integers, sample_steps within [0, steps], stride
and dtype checks); the scp/ssh invocations quote remote paths and
end-of-options the targets, and any transport failure (RemoteRunError,
timeout or OSError) degrades to a local NumPy run with a notice instead
of an exception. Configuration comes from PHONO_GPU_* variables,
optionally read from the gitignored .env (placeholder template
committed as .env.example).

Verified end to end against the reference host (RTX 4060, driver 570,
image cupy/cupy:v13.6.0): the 1200 x 1600 cell, 500 step benchmark
runs at 444 Mcell/s on the GPU versus 14.5 Mcell/s on local NumPy, a
30.7x stepping speedup with bit-identical returned frames; a stride 10
float32 job returns exactly the locally subsampled frames while cutting
the fetched frames archive from 4.84 MB to 27 kB.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added GPU-accelerated 2D acoustic FDTD engine with damping/material maps, sponge layers, obstacles, impedance boundaries, and plane-wave injection.
  * Added a job-based runner that executes locally (NumPy/CuPy autodetect), samples frames, and packages results.
  * Added remote GPU submission via SSH + Docker with automatic local fallback, plus an optional benchmark/verification mode.
* **Documentation**
  * Introduced a `.env.example` template for remote configuration.
* **Tests**
  * Added parity and round-trip verification tests covering GPU/CPU consistency, job pipeline behavior, and environment loading/fallback.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,7 +16,7 @@ PHONO_GPU_HOST=192.168.1.100
 PHONO_GPU_USER=gpuuser
 
 # Human label used in runner messages and benchmark reports.
-PHONO_GPU_NAME=lab GPU
+PHONO_GPU_NAME="lab GPU"
 
 # Docker image providing python3 + CuPy with CUDA support. The default is
 # verified against NVIDIA driver 570 on an RTX 4060.
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,26 @@
+# Remote GPU runner for the animation FDTD fields (scripts/fdtd_gpu_remote.py).
+# Copy to .env (gitignored) and adjust; real environment variables take
+# precedence over this file. Leave PHONO_GPU_HOST empty (or omit the file)
+# to run everything locally on NumPy.
+#
+# The host needs: SSH key access for PHONO_GPU_USER, Docker with the NVIDIA
+# runtime (docker run --gpus all) and enough disk in PHONO_GPU_WORKDIR for
+# per-job directories (created and removed automatically per run).
+#
+# The values below are placeholders; replace them with your own machine.
+
+# SSH host name or address of the CUDA machine.
+PHONO_GPU_HOST=192.168.1.100
+
+# SSH user on the host (key-based, no password prompts).
+PHONO_GPU_USER=gpuuser
+
+# Human label used in runner messages and benchmark reports.
+PHONO_GPU_NAME=lab GPU
+
+# Docker image providing python3 + CuPy with CUDA support. The default is
+# verified against NVIDIA driver 570 on an RTX 4060.
+PHONO_GPU_DOCKER_IMAGE=cupy/cupy:v13.6.0
+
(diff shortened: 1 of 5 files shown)


jmrplens:
@coderabbitai review

jmrplens:
Review addressed in 68b42219: the .env.example value with a space is quoted; build_job now validates cfl, damping, edge impedance and obstacle mask eagerly through the same helpers GpuFDTD2D uses (shared, so the contracts cannot drift) and stores the normalized values; sample_steps are deduplicated in build_job and defensively in run_job so frames always align with the reported steps; and the suite gained the submit local-fallback test, the load_env precedence test and eager-validation cases (25 passed).

jmrplens:
Second pass addressed: build_job now validates sponge_width bounds and sponge_reflection eagerly through a helper shared with GpuFDTD2D, with two new validation cases; and the exception tests hoist their array constructions out of pytest.raises so each block has a single throwing invocation (the six SonarCloud S5778 findings).

PR #420, opened 2026-07-29
Add the Spanish noise regulation RD 1367/2007 and the CTE DB-HR building-code indices

jmrplens:
Spain assesses activity noise with an index the ISO 1996 family does not define, and states its building-code insulation requirements in global quantities that are close relatives of, but not identical to, the ISO 717-1 weighted ratings. Both are now implemented from their legal texts, with the worked examples of the Spanish literature as numeric oracles.

## RD 1367/2007 (`environmental/spanish_regulation.py`)

The corrected level `LKeq,T = LAeq,T + Kt + Kf + Ki` of Annex I A.2 c, with the three reference procedures of Annex IV A.3.3:

- `tonal_correction` works on the unweighted one-third-octave spectrum, comparing each band against the **arithmetic mean** of its two neighbours, with the 8/12, 5/8 and 3/5 dB thresholds of the three band ranges and the largest emergent tone governing.
- `low_frequency_correction` and `impulsive_correction` grade `LCeq - LAeq` and `LAIeq - LAeq`.
- Each correction is 0, 3 or 6 dB and their sum is capped at 9 dB.

These are the regulation's own variants, not the ISO 1996 procedures the library already carries, and the module documents where each differs: the Annex C tonal adjustment is a critical-band method returning a continuous `Kt`, the Annex K survey method requires the band to exceed *both* neighbours by 15/8/5 dB and only flags prominence, and `impulsive_sound_adjustment` is the ISO/PAS 1996-3 onset-rate method on a calibrated signal. `Kf` has no ISO 1996 counterpart at all.

Around that:

- `NoisePhase`, `evaluation_period_level` and `long_term_corrected_level` integrate noise phases into the day, evening and night indices and into the annual `LK,x`, with the regulation's own rounding (`round_reported_level`: add 0,5 dB and take the integer part).
- The Annex II quality objectives (outdoor, indoor, vibration) and the Annex III immission limits (new transport infrastructure and its `LAmax` variant, activities and ports, and noise transmitted to acoustically adjacent premises) are exposed as lookups covering every row of every table.
- `assess_activity` applies the three criteria of Article 25 and returns an `ActivityAssessment` with `.plot()` and a one-page `.report()` inspection fiche.

The fiche renders in Spanish by default and follows the format real Spanish documents converge on rather than an invented one: the content that the municipal noise ordinances legally prescribe for an *acta de inspección acústica* (Annex VIII of the Sevilla and Almería ordinances, article 50 of the Madrid one), combined with the per-point one-page *ficha* geometry of accredited laboratory reports, in which each Article 25 criterion is shown against its own allowance.

## CTE DB-HR (`building/spanish_building_code.py`)

- The A-weighted global indices `RA`, `RA,tr`, `DnT,A` and `D2m,nT,Atr` computed by the direct Annex A formulae (A.5) to (A.7) over the **eighteen** one-third-octave bands 100 Hz to 5 kHz, with the four normalised source spectra of Tables A.2 to A.5 (pink noise, road traffic, railway, aircraft).
- Results expose `.value`, `.intermediate` (one decimal) and `.reported` (the integer), which is what clause 3.1.3.1 point 4 prescribes.
- Requirements: the facade table 2.1 keyed on `Ld` with the sheltered-facade and aircraft-noise rules, the airborne and shared-opening requirements of 2.1.1, the party-wall alternatives of 2.1.1 c), the impact limits of 2.1.2, the reverberation and absorption limits of 2.2, and the window-size correction of the Catálogo de Elementos Constructivos, checked through `check_db_hr_requirement` / `assess_db_hr`.

**The rounding convention is settled explicitly**, since this is the usual source of confusion. The module implements the direct Annex A route (the normative one for DB-HR) and the guide shows it agrees to the integer with the ISO 717-1 route of `weighted_rating_extended` on both spectra: the wall gives `R'A` 51,4 -> 51 dBA against `R'w + C100-5000` = 52 - 1 = 51, and the facade gives `D2m,nT,Atr` 32,8 -> 33 dBA against `D2m,nT,w + Ctr,100-5000` = 38 - 5 = 33.

The two rows are not equally strong evidence, and the guide says so. Ejemplo 7.1 publishes the wall's `R'w` 52, `C` -1 and the 51 dBA, so that row is a genuine external check; for the facade the book publishes only the 32,8 dBA, and the 38 and the -5 come from this library's own ISO 717-1 engine, so that row shows the two routes agreeing with each other rather than with a source. The guide also flags the naming trap: Annex H writes the terms as plain `C` and `Ctr` and defers to UNE-EN ISO 717-1 without narrowing the range, and the reading that they are the **enlarged-range** terms is the Manual's. It matters for `D2m,nT,A`, where the core-range `C` is -2 while `C100-5000` is -1; for the `D2m,nT,Atr` that DB-HR actually requires, `Ctr` and `Ctr,100-5000` are both -5 on this specimen, so that oracle does not discriminate between the two readings.

## Oracles

Every limit table is transcribed from the legal text itself and tested row by row (all six acoustic area types, all indoor and adjacent-premises rows, every `Ld` band and every column pairing of DB-HR table 2.1, both threshold boundaries of each correction).

The numbers come from *Manual de acústica ambiental y arquitectónica* (Avilés López & Perera Martín, Paraninfo):

| Oracle | Value | What it validates |
| :--- | :--- | :--- |
| Ejemplo 3.1 | `LKeq,d` 57 dB, `LKeq,e` 51 dB | phase corrections and period integration |
| Ejemplo 3.2 | `LK,d` 56 dB, `LK,e` 50 dB | the annual average over 303 operating days |
| Ejemplo 3.3 | fails as new, passes as existing | the three Article 25 criteria |
| Ejemplo 7.2 | `R'A` 51,4 dBA | the Annex A global index (pink noise) |
| Ejercicio 7.1 | `D2m,nT,Atr` 32,8 dBA | the same for road traffic |
| Ejemplo 7.4 | 26 -> 24 dBA | the window-size correction |
| Ejemplo 7.1 | `R'w` 52 dB, `C` -1, `Ctr` -5 | independent cross-check via the ISO 717-1 engine |

## Also in this PR

- Seven conformance checks and an errata entry for the `Kf` and `Ki` threshold tables of Annex IV A.3.3, whose 3 dB row prints the unsatisfiable condition "Si 10 > Lf <= 15"; the bracketing rows leave `10 < Lf <= 15` as the only consistent reading.
- Two figures and new guides in Spanish and English, Spanish-first for the regulation since that is the module's natural audience.

## Notes on what I corrected while finishing this

- The fiche defaulted to Spanish but **none of its 24 display strings had a Spanish translation**, so the "Spanish" fiche rendered entirely in English. All of them are translated now, along with the limit-row reference and description that the result carries in English.
- The fiche's assessment plot drew its three limit lines in the same colour and dash style, so the three legend entries were indistinguishable; they are now dotted, dashed and solid. The categorical period axis no longer inherits the 45-degree band-axis rotation.
- The DB-HR index plot put the band insulation (tens of dB positive) and the weighted transmitted level (tens of dB negative) on one axis, leaving half the figure empty; the transmitted level now has its own axis.
- The fiche overflowed to two pages once a third evaluation period was added, so the assessment figure now shrinks as the phase table grows.
- Added the DB-HR 2.1.1 c) party-wall requirement, which was missing, and recorded the residential/hospital condition that 2.1.1 b) ii attaches to its shared-opening variant.

## Not anchored on a published oracle

The Annex II Table B indoor quality objectives, the Table C vibration objectives, the Annex III Table A1/A2 infrastructure limits and the DB-HR clause 2 requirement tables are transcriptions of the printed legal tables, tested against those printed values; no worked example in the literature exercises them numerically. The `Kt` procedure has no published worked example either, so its tests drive the Annex IV A.3.3 table directly at each threshold boundary rather than reproducing a solved case.

## Summary by Sourcery

Implement Spanish environmental noise regulation RD 1367/2007 and Spanish building code CTE DB-HR indices end-to-end, exposing them through the public API, plots, conformance checks and guides.

New Features:
- Add RD 1367/2007 corrected level, tonal/low-frequency/impulsive corrections, evaluation-period integration, limit lookups and activity compliance assessment with plotting and PDF reporting.
- Add CTE DB-HR global indices (RA, RA,tr, DnT,A, D2m,nT,Atr) over the 100–5000 Hz bands, plus requirement lookups for facades, airborne/impact insulation and reverberation, with plotting and assessment helpers.

Enhancements:
- Expose the new Spanish regulation and building-code APIs, results and constants at the top-level environmental and building namespaces and in the global package __all__.
- Extend plotting modules with visualisations for RD 1367 tonal correction and activity assessment, and for DB-HR global indices and requirement checks.
- Update the conformance report to include new checks for RD 1367/2007 and CTE DB-HR, increasing the total passing checks and restructuring the room & building acoustics section.
- Improve report and plotting localisation so Spanish strings and limit-row descriptions render correctly in the RD 1367 inspection fiche and related figures.

Documentation:
- Add English and Spanish guides for the Spanish noise regulation (RD 1367/2007) detailing LKeq, corrections, periods, limits and compliance workflow.
- Add English and Spanish guides for the Spanish building code (CTE DB-HR) describing the Annex A indices, rounding, requirements and window-size correction.
- Update API reference, sidebar navigation, README and conformance docs (English and Spanish) to document the new modules, functions and additional conformance checks.
- Extend the errata documentation with a note on the RD 1367/2007 Kf/Ki threshold-table misprint and how the implementation interprets it.

Tests:
- Add regression tests for RD 1367 period/annual levels, corrections, limit tables and activity verdict, including coverage of the misprinted Kf/Ki thresholds.
- Add tests for the RD 1367 PDF inspection report to ensure single-page layout, localisation, escaping and that results and verdicts match the assessment.
- Add tests for the CTE DB-HR global indices, spectra, ISO 717 route agreement, window-size correction and all requirement helpers (facade, airborne, impact, reverberation).


jmrplens:
## Independent review

I re-derived every number in this PR from the primary sources rather than from the branch: the original BOE PDF of RD 1367/2007 (BOE-A-2007-18397, pp. 42952-42973) alongside the consolidated text, the DB-HR PDF, and chapters 3 and 7 of Avilés López & Perera Martín. I also recomputed both global-index oracles outside the library and ran mutation tests over the transcribed tables. Two things need fixing before this goes in; the rest is a list of smaller items and a note on what I confirmed.

---

## Blocking

### 1. The band arithmetic is wrong, and it is repeated in seven places

`src/phonometry/building/spanish_building_code.py:21`:

> the **eighteen** one-third-octave bands 100 Hz to 5 kHz, four more than ISO 717-1's sixteen.

Eighteen minus sixteen is two, not four. The extra bands are 4000 Hz and 5000 Hz, and only those two. The same sentence is in:

- `site/src/content/docs/guides/spanish-building-code.mdx:54-56`, where it also carries the follow-on "and those four high bands are the first reason the two routes do not always agree"
- `site/src/content/docs/es/guides/spanish-building-code.mdx:57` ("cuatro más que las dieciséis")
- `docs/spanish-building-code.md:30`
- `site/src/content/docs/reference/api/building/spanish-building-code.md:26`
- `llms-full.txt:28258`
- `site/public/llms/llms-sound-insulation.txt:2678`

This is the opening explanatory sentence of the whole DB-HR guide, on the one structural fact that separates the DB-HR route from ISO 717-1. It has to be right.

### 2. The railway variant returns the wrong index name

`spanish_building_code.py:398-415`, `d2m_nt_atr(..., spectrum="railway")` returns a result named `D2m,nT,Atr`. DB-HR says the opposite. Clause 3.1.3.4 point 1:

> Cuando el ruido exterior dominante es el ferroviario o el de estaciones ferroviarias, se debe usar la magnitud de aislamiento global D2m,nT,A. Cuando el ruido exterior dominante es el de automóviles o el de aeronaves, la magnitud del aislamiento global es D2m,nT,Atr.

And the Annex A definition of formula (A.5), which is the pink-noise `D2m,nT,A`, carries the railway case explicitly:

> En caso de que el ruido exterior dominante sea el ferroviario o el de estaciones ferroviarias también se utilizará este índice para la valoración global, pero usando los valores del espectro normalizado de ruido ferroviario.

Formula (A.6), the `D2m,nT,Atr` one, only picks up the aircraft case ("En caso de que el ruido exterior dominante sea el de aeronaves también se utilizará este índice"). Table H.1 prints the same split: Ferroviario -> `D2m,nT,A` (A.5); Automóviles / Aeronaves -> `D2m,nT,Atr` (A.6). So the aircraft handling is right and the railway handling is not: railway belongs to (A.5) under the name `D2m,nT,A`.

It happens to be numerically invisible today because Table A.4 is digit for digit Table A.3, which is exactly why it will stay wrong until someone prints a report with the wrong quantity name on it. The same mislabel is in:

- the guide table row `spanish-building-code.mdx:63` and its Spanish twin at line 66, where "facades with dominant rail noise" sits in a "Quantity it defines" column whose neighbouring row is `D2m,nT,Atr`
- `db_hr_facade_requirement(..., dominant_noise="railway")`, which returns `quantity="D2m,nT,Atr"`
- `test_aircraft_and_railway_facade_variants`, which asserts the two are numerically equal without asserting anything about the name

There is also no way to get a `D2m,nT,A` out of the public API other than `db_hr_global_index(values, "railway", name="D2m,nT,A")` by hand. A named helper would be worth having, since Table 2.1 is entered with a rail-dominant site often enough.

---

## Non-blocking

### 3. Annex H is credited with something it does not say

`spanish-building-code.mdx:113-116` says "Annex H of the document itself accepts the approximations $D_{nT,A} \approx D_{nT,w} + C_{100-5000}$ and $D_{2m,nT,Atr} \approx D_{2m,nT,w} + C_{tr,100-5000}$". Annex H writes (H.1) to (H.3) with plain `C` and `Ctr`, no range subscript, and then defers the definition to UNE-EN ISO 717-1 without narrowing the range. The enlarged-range reading is the Manual's, in the Nota on printed page 396 next to expressions [7.15] and [7.16]. The page then contradicts itself twelve lines later with "DB-HR calls them simply $C$ and $C_{tr}$". Attribute the enlarged-range reading to the Manual and keep Annex H's own wording.

While in there: Annex H also gives (H.2), `D2m,nT,w + C` as an approximation of `D2m,nT,A` for trains, which the guide omits and which is precisely the case in item 2.

### 4. The rounding claim holds, but the facade half of it is not an oracle

I ran both routes independently. Direct: `R'A` = 51.4432 -> 51.4 -> 51; `D2m,nT,Atr` = 32.8156 -> 32.8 -> 33. ISO 717-1: wall `R'w` 52, `C` -1, `Ctr` -5, `C100-5000` -1; facade `D2m,nT,w` 38, `C` -2, `Ctr` -5, `Ctr,100-5000` -5. Both agree to the integer, so the claim is true. Two caveats worth writing into the test docstring and the guide table:

- For the wall, `R'w` 52 and `C` -1 are printed in Ejemplo 7.1 and the library reproduces them, which is a genuine independent check. For the facade, the book prints neither `D2m,nT,w` = 38 nor `Ctr,100-5000` = -5; both come from `weighted_rating_extended`. The guide table at line 123 presents them in the same visual register as the published wall row.
- The "C is backwards" hazard is demonstrated on `C` vs `C100-5000` (-2 vs -1), that is on `D2m,nT,A`. For the facade index DB-HR actually requires, `Ctr` and `Ctr,100-5000` are both -5 on this specimen, so the oracle does not discriminate at all. The PR description's "if that is backwards every facade result is wrong" overstates what the data show.

As a bonus oracle the PR does not claim: `R'A,tr` on the same wall comes out 47.35 -> 47 dBA, which is the "47 dBA (= 52 - 5)" the book prints in Ejemplo 7.1. Worth adding as a third anchor, it costs one line.

### 5. One conformance row checks the library against itself

`scripts/conformance_report.py`, `_chk_dbhr_route_agreement` is titled "(Manual Ejemplo 7.1: 52 - 1 = 51 dBA)" but its `expected` is `weighted_rating_extended(...).rating + .c_100_5000`, not the printed 51. It is a self-consistency check wearing a published-oracle label. Set `expected` to the literal 51.0. The unit test already does this properly by asserting `rating == 52` and `c == -1` against the book.

### 6. One table boundary is not pinned

I mutation-tested 20 cells and thresholds across the four RD tables, the four DB-HR spectra, Table 2.1, the correction thresholds and the Article 25 allowances. Eighteen were caught. One real survivor: changing the last entry of `_FACADE_LD_BOUNDS` from 75.0 to 76.0 leaves the whole suite green, because the Table 2.1 tests only use Ld = 75.0 and Ld = 80.0. Add a case at 75.5 or 76.0. (The other survivor was a 0.05 m² nudge on the CEC 2.7 m² breakpoint, which is not meaningful.)

### 7. Public dataclasses accept states the code silently mishandles

- `DbHrRequirement.direction` is unvalidated. `check_db_hr_requirement` (lines 897-900) branches on `== "min"` and treats everything else as `"max"`, so a typo produces a confident, inverted verdict. The dataclass is exported.
- `NoisePhase.kt/kf/ki` accept any non-negative float. Annex IV A.3.3 grades each correction 0, 3 or 6 dB and nothing else, which is the whole point of the tables; 4.5 dB should not be silently accepted.

### 8. A new activity with no annual data reports PASS

`assess_activity` defaults to `new_activity=True`. With no `long_term_levels` and no `operating_days`, `long_term_pass` is `None`, `PeriodAssessment.complies` treats `None` as satisfied, and the assessment returns `complies=True` while never evaluating Article 25.1 b i, which is mandatory for a new activity. The fiche then prints an empty cell in that column and a PASS verdict row. `test_assess_activity_without_annual_information_skips_that_criterion` pins this. Either require annual data when `new_activity=True`, or make the verdict say "not fully assessed".

### 9. The governing period can be a period that passes

`_report/rd1367.py:288`: `max(result.periods, key=lambda p: p.reported_level - p.limit)` ranks on the daily criterion only. In a case that fails on the annual criterion alone, the boxed conclusion can name a compliant period as the governing one.

### 10. Smaller things

- `RD1367_PERIOD_HOURS`, `RD1367_PERIOD_CLOCK_LIMITS`, `ACOUSTIC_AREA_TYPES` and `DB_HR_NORMALISED_SPECTRA` are plain mutable dicts in the public API. I confirmed that assigning to `DB_HR_NORMALISED_SPECTRA["pink"]` changes what `ra()` returns process-wide, and that assigning to `RD1367_PERIOD_HOURS["day"]` changes what `assess_activity` accepts. `MappingProxyType` closes it.
- Result objects alias the caller's arrays: `tonal_correction(levels, freqs).levels is levels` and `ra(values).band_values is values` when a float64 array is passed, so mutating the input afterwards silently rewrites the stored result.
- `"sanitary"` means two different things across the two new modules, and both are exported into the same top-level namespace: in `spanish_regulation` it is an alias of *hospitalario* (Annex II Table B), while in `spanish_building_code` it is the non-hospital ambulatory use of Table 2.1 footnote (1), deliberately distinct from `"hospital"`. Worth a one-line warning in each docstring.
- `assess_activity` only validates `year_days` when `operating_days` is given, and `int(year_days)` truncates a float silently.
- `_report/rd1367.py:77` imports `ReportMetadata` at module level although with `from __future__ import annotations` it is only used in annotations.

---

## Verified correct

Digit by digit against the legal texts, not sampled:

- **Annex II Table A** as amended by RD 1038/2012: e 60/60/50, a 65/65/55, d 70/70/65, c 73/73/63, b 75/75/65, including the counter-intuitive c-night 63 sitting below d-night 65. **Table B** all six rows. **Table C** 75/72/72.
- **Annex III** A1 (55/60/65/68/70 day-evening with nights 45/50/55/58/60), A2 (80/85/88/90/90), B1, and B2 all eight rows. Cross-checked a second time against Tablas 3.7 and 3.8 of the Manual, which reproduce B1 and B2.
- **Annex IV A.3.3 Kt**: the three band ranges 20-125, 160-400 and 500-10000 Hz with 8/12, 5/8 and 3/5 dB, `Ls` as the arithmetic mean of the two adjacent bands, unweighted spectrum, largest emergent tone governing, end bands and out-of-table bands excluded. The three ranges tile contiguously with no gap and no overlap under the 6 % centre tolerance.
- **Kf and Ki**: the RD defines both as broadband differences (`LCeq,Ti - LAeq,Ti` and `LAIeq,Ti - LAeq,Ti`) with no band ranges at all, and the implementation correctly has none. Thresholds `<= 10` / `(10, 15]` / `> 15` and the 9 dB cap on the sum.
- **Article 25** 1 b i/ii/iii and 25.2, and Article 14.2's 5 dB reduction with the "existing urbanised area" definition of Article 3 b (before entry into force, 24 October 2007).
- `round_reported_level` = `floor(x + 0.5)`, matching "se redondeará incrementándolo en 0,5 dB(A), tomando la parte entera como valor resultante" of Annex IV A.3.4.2 a and b.
- **DB-HR Annex A** Tables A.2, A.3, A.4 and A.5, all 72 values, including the 4000 Hz pink value printed as "-10" and the fact that A.4 is identical to A.3. Formulae (A.5), (A.6) and (A.7) mapped to the right spectra.
- **Table 2.1**: the five Ld bands, the two series 30/32/37/42/47 and 30/30/32/37/42, and all twelve use and room pairings including the dormitorios/estancias to estancias/aulas transposition between the two blocks. Sheltered facade Ld - 10, aircraft +4, default 60 dBA.
- **2.1.1** a) i-iv, b) i-iii and c), including all three shared-opening variants (30/50, 20/50, 30/50), the residential-or-hospital condition attached to b) ii, and the correct absence of a shared-opening variant for a protected room against a plant or activity room. **2.1.2** 65/60/60. **2.2** 0,7 / 0,5 / 0,9 s and A/V >= 0,2 m²/m³.
- The CEC window-size table and Ejemplo 7.4 (4 m², 26 -> 24 dBA), against Tabla 7.3 on printed page 408.

On the oracles specifically, I opened each worked example and checked inputs as well as outputs:

- **Ejemplo 3.1** (printed 171-172): the two phase rows are the book's (50 dB with Kt 6, Kf 3, Ki 0 -> 59; 48 dB with Kt 3, Kf 3, Ki 0 -> 54), the phase split is the book's (2 + 6 + 4 h and 2 + 2 h with the closed phase at 0 dB), and 57 dB and 51 dB are what the book prints. Nothing reverse-engineered.
- **Ejemplo 3.2** (printed 172): 303 open days and 62 closed -> 56 dB and 50 dB. Worth noting that the book feeds the *rounded* 57 and 51 into the annual average, which is exactly what `assess_activity` does, so the pipeline matches the source and not just the endpoint.
- **Ejemplo 3.3** (printed 175): area type a, criteria 60 / 58 / 55, fails as new on `LK,d` = 56 and passes under Article 25.2. Reproduced.
- **Ejemplo 7.2 and Ejercicio 7.1**: I recomputed both sums by hand outside the library and got 51.4432 and 32.8156 against the printed 51,4 and 32,8. The full input spectra in the tests are the book's tables.
- **Ejemplo 7.5 / Ejercicio 7.2** figures quoted in the guide (31.53, 31.97, 35.63 dBA) reproduce too.

On the **errata entry**: I did not take the consolidated HTML at face value, because a broken inequality is exactly the kind of thing a consolidation pipeline invents. I pulled the original BOE PDF of 23 October 2007 and the enacted text really does print "Si 10 >Lf <=15" and "Si 10 >Li <=15". So the defect is in the enacted regulation, not in the BOE's consolidation, the characterisation in `docs/ERRATA.md` is accurate, and the supporting claim that the derived autonomous and municipal texts print `10 < Lf <= 15` checks out as well. Good entry.

House rules: no em-dashes in the new prose, no `assert` in the new `src/` files, no module-level level-2 relative imports in `_plot/` or `_report/`, `.plot()` on all four new result objects, and EN/ES parity is 40/40 and 34/34 on headings, tables and snippets with identical numeric token sequences on both pages. Gates all green in a clean worktree: ruff, ruff format, `mypy src scripts` (245 files), bandit, 182 tests, `test_package_architecture` (22), `check_api_reference` (974 names). The 182 tests come from 84 test functions through parametrisation, so the count is not padded.

jmrplens:
Thank you, this is an unusually careful review. I went back to the primary sources for every point rather than taking the findings on trust, and I am pushing fixes for both blocking items plus most of the rest. Point by point.

## Blocking

### 1. Band arithmetic: confirmed, fixed

Eighteen minus sixteen is two. The extra bands are 4 kHz and 5 kHz and nothing else. Fixed at the hand-written sources (the module docstring and the three guide files: EN, ES and the `docs/` mirror) and then regenerated, so the API page and the two `llms` files pick it up rather than being edited in place. The follow-on clause now reads "the 4 kHz and 5 kHz bands, which are the first reason the two routes do not always agree".

### 2. Railway index name: confirmed, fixed

You are right, and this was the most valuable finding in the review. I read it out of the DB-HR document myself rather than relying on the quoted text. Three independent places say the same thing:

- The Annex A definition that introduces (A.5) is headed "Diferencia de niveles estandarizada, ponderada A, en fachadas, en cubiertas y en suelos en contacto con el aire exterior, **D2m,nT,A**: ... para ruido rosa", and the note that follows says the railway case "también se utilizará **este índice**" with the Table A.4 spectrum. So railway attaches to (A.5), which is `D2m,nT,A`.
- (A.6) is headed `D2m,nT,Atr` and its note picks up only the aircraft case.
- Clause 3.1.3.4 point 1 states it outright, and Table H.1 tabulates the same split: Ferroviario to `D2m,nT,A` via (A.5), Automóviles and Aeronaves to `D2m,nT,Atr` via (A.6).

So the library was putting the wrong quantity name on a rail-dominant facade. What changed:

- Added `d2m_nt_a()` for Formula (A.5), accepting `"pink"` (default) or `"railway"`, and narrowed `d2m_nt_atr()` to `"traffic"` and `"aircraft"`. Each refuses the other's spectra with a message naming the right function, so the failure mode is an exception rather than a silently mislabelled report.
- `db_hr_facade_requirement(dominant_noise="railway")` now returns `quantity="D2m,nT,A"`; the numeric limit from Table 2.1 is unchanged, as it should be.
- The guide tables in both languages carry a Formula column and name the quantity per row, with a paragraph explaining exactly the trap you identified: because Table A.4 is digit for digit Table A.3, a rail-dominant facade comes out at the same number but not the same quantity.
- `test_aircraft_and_railway_facade_variants` is replaced by tests that assert on the *name*, which is the only thing distinguishing the two, plus tests that each helper rejects the other's spectra and that the requirement is stated in the right quantity.

## Non-blocking

### 3. Annex H attribution: confirmed, fixed

Annex H writes (H.1) to (H.3) with plain `C` and `Ctr` and defers to UNE-EN ISO 717-1 without narrowing the range, exactly as you say. The enlarged-range reading is the Manual's, in the note to expressions [7.15] and [7.16]. Both guides now print all three approximations as Annex H writes them, including **(H.2), `D2m,nT,w + C` for trains**, which I had omitted and which is precisely the case in item 2, and then attribute the enlarged-range reading to the Manual. The self-contradiction twelve lines later is gone.

### 4. The facade half is not an oracle: confirmed, fixed

Agreed, and this was a real overstatement on my part. The guide table now carries a "Published?" column: the wall row is marked published (Ejemplo 7.1 prints `R'w` 52, `C` -1 and the 51 dBA) and the facade row is marked not published (the book prints neither the 38 nor the -5, both of which come from this library's own ISO 717-1 engine). The prose says plainly that the second row shows the two routes agreeing with each other rather than with a source, and the test docstring says the same.

I also took your bonus oracle. `R'A,tr` on that wall is 47.35 -> 47 dBA, which is the 47 dBA the book prints, so it is now a third anchor with its own test and its own conformance row.

On the "if that is backwards every facade result is wrong" line in the PR description: you are right that the data do not support it. On this specimen `Ctr` and `Ctr,100-5000` are both -5, so the oracle does not discriminate for `D2m,nT,Atr` at all; only the `C` versus `C100-5000` pair (-2 versus -1) discriminates, and that is `D2m,nT,A`. The guide now says exactly that and I have corrected the PR description.

### 5. Conformance row checking the library against itself: confirmed, fixed

Fair, and worth being strict about. `expected` is now the literal `51.0` printed in Ejemplo 7.1, and the row is retitled to name the Manual rather than "Annex A vs ISO 717-1". The comment in the check says why. The unit test keeps the genuine `R'w == 52` and `C == -1` assertions against the book.

### 6. Unpinned Table 2.1 boundary: confirmed, fixed

Good catch, and mutation testing was the right way to find it. The parametrisation now walks both sides of every band edge (60.0/60.1, 65.0/65.1, 70.0/70.1, 75.0/75.1, plus 76.0), so moving the last bound to 76.0 now fails. I agree the 0.05 m2 nudge on the 2.7 m2 CEC breakpoint is not worth pinning.

### 7. Unvalidated public dataclass fields: confirmed, fixed

Both fixed.

- `DbHrRequirement` validates `direction` in `__post_init__`, so a typo raises instead of silently inverting the verdict.
- The three corrections are graded against the Annex IV A.3.3 tables: each must be 0, 3 or 6 dB, exposed as `RD1367_CORRECTION_VALUES`. 4.5 dB raises. Enforced in `total_correction`, so it covers `NoisePhase`, `corrected_level` and the phase constructor uniformly.

### 8. New activity with no annual data: confirmed, fixed

You are right that this was the worst of the smaller items, because it returns a confident PASS for a criterion it never evaluated. I took the first of your two options: `assess_activity` raises when `new_activity=True` and neither `long_term_levels` nor `operating_days` is given, with a message pointing at `new_activity=False` for an Article 25.2 inspection. The guard runs after the structural validation so a malformed measurement set still reports its own more specific error. The test that pinned the old behaviour is replaced by one asserting the refusal and one asserting that an in-operation inspection is still assessed without annual data.

### 9. Governing period: confirmed, fixed

The boxed conclusion now ranks periods on the worst margin across every criterion that was actually evaluated, not the daily one alone.

### 10. Smaller things

- **Mutable public dicts**: `DB_HR_NORMALISED_SPECTRA`, `RD1367_PERIOD_HOURS`, `RD1367_PERIOD_CLOCK_LIMITS` and `ACOUSTIC_AREA_TYPES` are now `MappingProxyType`, with tests asserting they reject assignment.
- **Array aliasing**: `tonal_correction` and `db_hr_global_index` copy, with tests that mutate the input afterwards and assert the stored result is unchanged.
- **`"sanitary"` collision**: both module docstrings carry an explicit note that the name means the non-hospital ambulatory use of Table 2.1 footnote (1) in one module and *hospitalario* in the other, and that both reach the top-level namespace.
- **`year_days`**: validated whichever annual input is given, and a fractional count of days is rejected rather than truncated.
- **`ReportMetadata` import**: moved under `TYPE_CHECKING`.

## On the verification section

Thank you for pulling the original BOE PDF of 23 October 2007 for the errata entry rather than trusting the consolidated HTML. That was the right instinct and it is the one check I could not have made as convincingly from the consolidated text alone, since a broken inequality is exactly what a consolidation pipeline might introduce. Good to have the enacted text confirm it.

Local gate after these changes: ruff, `mypy src scripts` (245 files), bandit, `check_api_reference` (976 names now), i18n parity 140/140, all 1568 figures matching, 437/437 conformance checks, and 1663 tests green across the affected suites.

PR #421, opened 2026-07-29
IEC 61043 instrument-class verification and the ISO 9614-1 F1 field indicator

jmrplens:
Grades a two-microphone sound-intensity chain against IEC 61043 Table 2, and
completes the ISO 9614-1 Annex A field indicators with F1.

## Instrument class verification (new `metrology/intensity_compliance.py`)

- `residual_index_limits(device, spacing=..., frequencies=...)` returns the
  Table 2 minimum pressure-residual intensity index per one-third-octave band
  from 50 Hz to 6.3 kHz for a **probe**, a **processor** or a **complete
  instrument**, in class 1 and class 2, rescaled to any microphone separation
  by the Note 1 rule `+10 lg(x/25)`.
- `verify_intensity_class` / `intensity_class_compliance` grade a measured
  `δpI0` spectrum against both masks band by band: the class the chain meets
  (the loosest class every band clears, or `None`), the per-band margins, the
  bands that block the next class up, and a `range_limited` flag when the
  verified bands cover neither frequency range of clause 6.1. The octave-band
  alternative of clause 6.1 is open to a processor and to the instrument built
  on one, but not to a probe, whose index clause 12.4 determines at
  one-third-octave intervals over the whole 50 Hz to 6.3 kHz range.
- `instrument_class_from_components` applies the clause 8 rule for a probe and
  a processor supplied separately.
- The result exposes `.plot()` (measured index over the two Table 2 masks, the
  pass region of the achieved class shaded, the blocking bands ringed) and
  `.report()`, a one-page verification fiche.
- `phase_mismatch_from_residual_index` and `residual_index_from_phase_mismatch`
  convert between the index and the equivalent channel phase mismatch,
  `δpI0 = 10 lg(kd/φs)`, in both directions.

## ISO 9614-1 field indicator F1

`temporal_variability_indicator` implements equations (A.1)-(A.2): the
coefficient of variation of the M short-time-averaged normal-intensity samples
taken at one typical position in the initial test (clause 8.2), evaluated in
every measurement band. `field_indicators` accepts those samples as
`temporal_intensity` and fills `f1` alongside F2, F3 and F4;
`FieldIndicators.field_is_stationary()` checks it against the Table B.3 limit
of 0.6 (`TEMPORAL_VARIABILITY_LIMIT`), and the indicator plot draws F1 and that
limit on the dimensionless twin axis beside F4.

## Oracles

| Piece | Oracle |
| :--- | :--- |
| Table 2, all 22 bands x 3 device kinds x 2 classes | Table 2 of EN 61043:1994 (standard page 14), transcribed digit for digit, cross-checked against its reproduction as Table 6.1 in Fahy, *Sound Intensity* 2nd ed. (printed p. 136). The two sources agree exactly. |
| Table 2 internal consistency | The instrument column is the probe and processor residual intensities added as powers, `10^(-d_i/10) = 10^(-d_p/10) + 10^(-d_q/10)`. Every row satisfies it within the rounding of the printed figures, so a mistyped digit anywhere would show up. |
| Note 1 separation rule | Exact closed form, checked at 2.5, 6, 12.5, 25, 50 and 250 mm. |
| Clauses 6.1, 8, 12.4 | The clause text of EN 61043:1994. |
| `δpI0` to phase mismatch | Fahy 6.8 (printed p. 135): 20 dB at 1 kHz over 25 mm is a mismatch of a hundredth of `kd`, about 0.26 degrees. Plus exact round-trip, the 10 dB/decade slope and the `10 lg 2` spacer identity. |
| F1 | The closed form of equation (A.1) and identity checks: F1 is exactly the statistic F4 (A.8) computes, taken over the M time samples instead of the N positions; scale invariance; the two-sample reduction; a case landing exactly on the Table B.3 threshold. |

Four new conformance rows (Table 2, Note 1, the Fahy phase check, F1), a new
figure `intensity_class` in both languages and themes, and two new sections in
the intensity guide in English and Spanish.

## Not anchored on a published numeric example

**F1.** ISO 9614-1 gives no worked example for F1 to F4, and the real cases in
the literature present the indicators only as figures. F1 is therefore anchored
on the definition itself and on closed-form identity checks, as described
above, not on a published number. Stating that plainly rather than manufacturing
an oracle.

The residual-intensity *test* of IEC 61043 is still out of scope: `δpI0` is a
value the caller measures on their own probe-analyser chain, with the spacer
that will be fitted, and supplies. What is new here is the grading of it. The
non-real-time class 2X of clause 4 is not modelled.

## Summary by Sourcery

Add IEC 61043 sound-intensity instrument class verification utilities and ISO 9614-1 F1 temporal variability support, with plotting, reporting, tests, and documentation updates.

New Features:
- Introduce metrology.intensity_compliance module to grade p-p sound-intensity chains against IEC 61043 Table 2, including result object with plotting and PDF report generation.
- Expose residual index and phase-mismatch conversion helpers and new temporal variability indicator F1 API, including constants and helpers on FieldIndicators.
- Extend field_indicators to carry F1 samples and provide a stationarity check against ISO 9614-1 Table B.3.

Enhancements:
- Add plotting support for IEC 61043 residual index masks and integrate new intensity class figure generation into docs and report generation tooling.
- Update English and Spanish intensity guides, API reference, and conformance docs to describe instrument-class verification and F1 usage.
- Extend i18n mappings and API taxonomy to cover new intensity compliance and reporting features.

Documentation:
- Document new IEC 61043 class-verification and ISO 9614-1 F1 support in guides, API reference, changelog, and landing pages, including new intensity_class figures in both languages.

Tests:
- Add unit tests for intensity_compliance logic, report rendering, temporal variability indicator behaviour, and updated plotting/i18n paths.
- Expand conformance_report with new oracles for IEC 61043 Table 2, separation rule, phase-mismatch check, and ISO 9614-1 F1 identity.

## Summary by CodeRabbit

- **New Features**
  - Added ISO 9614-1 F1 temporal variability calculations and stationarity checks.
  - Added IEC 61043 instrument-class verification for two-microphone sound-intensity chains, including spacing adjustments, component grading, phase-mismatch conversions, plots, and PDF reports.
  - Added range-limited compliance reporting and Spanish-language visualization support.

- **Documentation**
  - Expanded intensity guidance, API references, errata, and examples.
  - Updated conformance reporting to 459 passing checks across 303 standards, with four additional intensity-related checks.


jmrplens:
@coderabbitai review

jmrplens:
Thanks, this was a useful pass. All of it is addressed in the latest push except one point I am deliberately leaving alone.

**Fixed**

- **Octave-only class 1 was over-certified.** This was a real bug and the most valuable catch. Clause 6.1 gives the one-third-octave range to class 1 and offers the octave range only to class 2, so `range_limited` now accepts the octave set only when the verdict is class 2. The test that pinned the old behaviour has been split into `test_octave_band_subset_attests_a_full_range_class_2` and `test_octave_band_class_1_verdict_is_still_range_limited`, covering both outcomes.
- **Edition identifier.** Correct: the international publication is IEC 61043:1993 and EN 61043:1994 is the European adoption. Fixed throughout the new module, the fiche, the conformance rows and the tests, including the pre-existing `IEC 61043:1994 Clause 5` row and the `emission/intensity.py` docstrings that carried the same conflation. References to `EN 61043:1994` are left as they are, since those cite the European document whose printed page the table was transcribed from.
- **Markdown table pipes.** The conformance row now reads `max absolute deviation ... dB`, so the six-column table renders correctly in `docs/CONFORMANCE.md` and both transplanted site copies.
- **Undefined `measured_delta_pi0`.** Both guides now bind an explicit 22-band placeholder spectrum before the call, and the snippet reproduces the class 2 verdict and the 100 Hz band that the surrounding prose describes.
- **Non-finite intensity samples** are rejected before the positivity test, so a NaN no longer propagates into a silent NaN F1/F4.
- **Scalar F1 for a one-band surface.** F1 now follows the shape of its siblings: scalar for a 1D per-position surface (so `field_is_stationary()` returns `True`, not `array([True])`), and a length-one array for a 2D one-band surface, where `f2`/`f3`/`f4` are arrays too. Regression test added for both directions.
- **Decimal separators in the English fiche strings.** `1.2048 kg/m³` and `6.3 kHz` in the English templates, with the Spanish values keeping commas. The docstrings keep commas, which is the existing house style in this package.

**Not changing: "accredited" in the result docstrings**

The phrases flagged in `intensity_compliance.py` are verbatim the wording its sibling `metrology/compliance.py` already uses for the IEC 61260 filter fiche ("Writes a one-page accredited report:", "render an accredited `.report()` fiche"), and the same convention runs through `_report/` and `building/insulation.py`. Diverging in this one module would make the API reference inconsistent for no functional gain, and the rendered fiche carries the "results relate only to the tested specimen" disclaimer plus the new scope limiter stating that the residual index is one IEC 61043 requirement among others. Renaming the convention is worth doing repository-wide or not at all, and that is out of scope here.

jmrplens:
Adversarial review against the primary sources (UNE-EN 61043:1999 and BS EN 61043:1994 texts, Fahy 2e Table 6.1 / section 6.8 / eqn (7.16), UNE-EN ISO 9614-1:2010 Annex A, Annex B and Table B.3). I re-derived the numbers, re-rendered the fiche, and mutation-tested the oracle tables. Verdict up front: **no blocking defects**. A handful of non-blocking items below.

## Blocking

None found.

## Non-blocking

1. **Clause 6.1 wording follows a defect in the UNE translation.** `src/phonometry/metrology/intensity_compliance.py` (module docstring, "a class 2 processor covers at least 45 Hz to 5,6 kHz in octave bands", repeated in the `verify_intensity_class` docstring and in both guides) matches UNE-EN 61043:1999, but the EN/IEC English text reads: "Class 2 processors shall, at least, cover the range from 45 Hz to 7,1 kHz in one-third octave bands, **or** the range from 45 Hz to 5,6 kHz in one octave bands." The Spanish translation dropped the first alternative. The *code* is unaffected (full third-octave coverage attests the full range for either class, which is exactly the IEC reading), so this is prose only: the octave range should be described as the class 2 *alternative*, not the class 2 range. Also a candidate for `docs/ERRATA.md`: the UNE-EN 61043:1999 translation of 6.1 omits half the sentence.

2. **Wrong clause cited when F1 rejects a non-positive mean.** `_coefficient_of_variation` (src/phonometry/emission/intensity.py) raises "the test conditions do not satisfy ISO 9614-1 (clause A.2.3)" for both F4 and F1. For F4 that citation is right (A.2.3 is where the standard says a negative surface sum invalidates the test). For F1 it is not: A.2.1 places no positivity condition on the M temporal samples at one position, and A.2.3 talks about the surface scan, not the initial test. Rejecting a non-positive temporal mean is a defensible design choice (F1 as defined would come out negative and the Table B.3 criterion would be vacuous), but the message should not attribute it to A.2.3 when reached through `temporal_variability_indicator`.

3. **ES guide annotates printed outputs with decimal commas.** `site/src/content/docs/es/guides/intensity.mdx` lines ~201, ~457 and ~460 comment `print(round(f1, 3))` with `# 0,177`, and similarly `# 0,26 grados` and `# 24,0 dB`. Python prints `0.177`, `0.26`, `24.0`. The other ES guides annotate outputs with the true printed form (e.g. `diffusers.mdx` `# 0.1343`, `absorption-measurement.mdx` `# 0.05`); these three should follow that convention, keeping the comma only in surrounding prose.

4. **Note 9 qualifier dropped in the guides.** ISO 9614-1 Note 9 recommends the 8 s to 12 s short averaging time *for periodic signals* ("Para señales periódicas, se recomienda un tiempo corto de promediación de entre 8 s y 12 s, o cualquier número entero de ciclos"). The module docstring keeps the qualifier; `docs/intensity.md` and both `intensity.mdx` guides say "Note 9 suggests M = 10 and 8 s to 12 s per sample", which presents it as unconditional.

5. **`range_limited` prose corner for a failed chain over the octave set.** `verify_intensity_class` grants full-range status to the 7 octave bands only when `overall == 2`, so a chain that meets *no* class over exactly those bands is flagged `range_limited`, and the fiche then prints "the stated class attests the bands listed above..." directly under the "Does not comply" box (src/phonometry/_report/iec61043.py:348-361). The flag itself is defensible; the fiche sentence just reads oddly when there is no stated class. Cosmetic.

## Verified correct

- **Table 2 / Fahy Table 6.1, digit for digit, three ways**: `_TABLE2` in `intensity_compliance.py`, and independently `IEC61043_TABLE2` in `tests/reference_data.py`, agree cell-for-cell with UNE-EN 61043:1999 (Tabla 2, page 14), with the BS EN 61043:1994 English text, and with Fahy Table 6.1 (printed p.136). All 132 cells, including the irregular 315/400/500/630 Hz instrument class 2 steps and the lone 14,5 at 400 Hz. The power-sum structural test (probe + processor = instrument within print rounding) is a genuinely independent cross-check and holds in every row.
- **Mutation hardness**: I corrupted single cells at 315, 400, 50 and 6300 Hz across different columns, flipped the Note 1 sign, changed the limit comparison from >= to >, changed the phase exponent from -δ/10 to -δ/20, changed F1's N-1 to N and the 0,6 limit to 0,5. Every mutation was caught by the new tests.
- **Note 1 direction and application point**: +10 lg(x/25) is added to the tabulated minima (the limits), exactly as the standard and Fahy state, and the exact +10 dB at 250 mm and +3,0103 dB at 50 mm come out right.
- **Phase conversion**: Fahy eqn (7.16) (checked visually on printed p.147, since the equations are images in the scan) is φs = ±(kd)·10^(-δpI0/10); the implementation matches, kd in degrees = 360·f·d/c, round trip exact, and the 0,26 degrees at 20 dB/1 kHz/25 mm reproduces. Annex D of the standard independently confirms the form (0,79 degrees for the 12 mm, 1 kHz, δ = 12 dB case).
- **F1**: equation (A.1)/(A.2) reproduced exactly (sample s.d. with M-1, normalised by the algebraic mean), Table B.3 criterion F1 > 0,6 with the limit itself passing, evaluated per band as A.1 requires, and B.1.4 (before/after each surface) correctly cited. F2, F3, F4 and Ld = δpI0 - K (K = 10/10/7 per Table 1) are undisturbed; the F4 refactor through the shared helper is behaviour-preserving.
- **Verdict logic**: inclusive limits match "equal to, or higher than" (6.9), overall class is the loosest per-band class, one failing band demotes the chain, out-of-table frequencies raise instead of extrapolating, and the clause 8 combination rule (only 1+1 gives class 1, 2X rejected) matches the standard's enumeration.
- **Fiche**: one page in EN and ES, bare and with metadata, compliant and non-compliant; pass-region fill is opaque pale with `edgecolor="none"`, `zorder=0` (no `alpha=` fills in the new renderer); margins printed with the same convention as the IEC 61260 fiche.
- **House rules**: `.plot()` on the new result, no `assert` in src/, no module-level level-2 relative imports in `_plot/`/`_report/`, all nine new `__all__` names documented in `docs/api-reference.md`, EN/ES guides structurally in parity, and every snippet output in the docs reproduces (`0.177`, class 2 with 100 Hz failing, -3,19 dB offset, 0,26 degrees, 24,0 dB).

jmrplens:
Thanks, this is a useful pass. I verified all five against the primary sources before changing anything. All five stand and I have taken all five; nothing to push back on. Details below.

## 1. Clause 6.1 wording follows a defect in the UNE translation

Accepted, and the omission is now in the errata registry.

Confirmed against both prints. UNE-EN 61043 (April 1999), clause 6.1, is a single sentence:

> Los procesadores de clase 2 deberán cubrir, al menos, el rango desde 45 Hz a 5,6 kHz en bandas de octava.

BS EN 61043:1994, same clause:

> Class 2 processors shall, at least, cover the range from 45 Hz to 7,1 kHz in one-third octave bands, or the range from 45 Hz to 5,6 kHz in one octave bands.

The class 1 sentence is equivalent in the two documents, so the divergence is confined to the class 2 sentence, and the Spanish document declares itself "la versión oficial, en español, de la Norma Europea EN 61043 de enero 1994, que a su vez adopta la Norma Internacional CEI 61043:1993". Two pieces of internal evidence put the English reading beyond doubt, and both sit inside the Spanish print itself: its Table 2 tabulates the pressure-residual intensity index for class 2 processors at all 22 one-third-octave centres, and its Note 2, translated faithfully ("Para procesadores con análisis en bandas de octavas únicamente, los requisitos se aplican únicamente a las frecuencias centrales de las bandas de octava"), carves out octave-only processors as a special case. Both are redundant if every class 2 processor is by definition an octave-band one.

On whether it belongs in `docs/ERRATA.md`: yes. The stated scope covers "standards (ISO, IEC, EN)", entries are keyed to "the specific printed editions cited", and dropping a normative alternative from an official national version is a defect of a published document rather than a property of the method, which is what the file's closing section is reserved for. New entry "UNE-EN 61043:1999, clause 6.1 (class 2 frequency range dropped in translation)", quoting both readings verbatim as above, status unreported (national translation, not the issuing body's text).

The prose is corrected in the module docstring, in the `verify_intensity_class` docstring, in `docs/intensity.md` and in both guides: the one-third-octave range is required for class 1 and available to class 2, and the octave range is offered to class 2 as an alternative. The guides and the module docstring now also state which text the library follows and point at the errata entry. No code change, as you say: the existing `full_range` rule was already the IEC reading.

## 2. Wrong clause cited when F1 rejects a non-positive mean

Accepted, right on both halves. A.2.3 ends with "Si sum(Ini/I0) es negativo en alguna banda de frecuencia, las condiciones del ensayo no satisfacen los requerimientos de esta parte de la Norma ISO 9614 en esa banda", so F4 and F3 may attribute the rejection to the standard. A.2.1 defines F1 over the M short-time samples at one position and states no positivity condition at all.

The shared `_coefficient_of_variation` now takes the message from its caller. F4 keeps the A.2.3 citation. F1 says the mean is not positive and that the indicator is therefore not meaningful at that position, without claiming any standard backing, and its docstring records that the rejection is this library's choice rather than the standard's: F1 as defined would come out negative and the Table B.3 criterion would be vacuous. New test `test_non_positive_mean_cites_a23_only_for_the_surface_indicators` pins that the surface message contains A.2.3 and the temporal one does not.

## 3. ES guide annotates printed outputs with decimal commas

Accepted. I checked the convention rather than assuming it: across the Spanish guides, `print(...)` lines annotated with a numeric literal run 219 with a decimal point against 7 with a comma, and three of those seven were mine. They now read `# 0.177`, `# 0.26 grados` and `# 24.0 dB`. I left `# True (límite 0,6 de la tabla B.3)` as it is, since the comma there is prose quoting the standard's limit and not a printed value.

For the record, the four remaining stragglers are in `microphones.mdx`, `sound-quality.mdx` and `underwater-propagation.mdx`. Out of scope for this branch, but worth a sweep later.

## 4. Note 9 qualifier dropped in the guides

Accepted. Note 9 reads "Normalmente M tomará un valor de 10. Para señales periódicas, se recomienda un tiempo corto de promediación de entre 8 s y 12 s, o cualquier número entero de ciclos", so M = 10 is unconditional and the 8 s to 12 s window is not. All three guides now read "Note 9 suggests M = 10, and, for periodic signals, 8 s to 12 s per sample or any whole number of cycles", which matches the module docstring that already had it right.

## 5. `range_limited` prose corner under a "Does not comply" box

Accepted, and reproduced: a processor evaluated over exactly the 7 octave bands that meets no class comes out `range_limited`, because full-range status over that set is granted only at `overall == 2`, and the strip then discusses what "the stated class attests" with nothing to refer to.

The flag stays as it is, for the reason you give. The fiche now selects the wording from `overall_class`: with no class reached it reads "this verification covers the bands listed above and not the full frequency range of the standard". The Spanish string is added alongside, and a new test renders that case in both languages and asserts the swap in each.

## Housekeeping

Rebased onto current main (#424, #427, #428, #429, #430, #432). The conflicts were all in generated or count-carrying artefacts plus one `__all__` line; I took main's side and regenerated rather than hand-merging, then updated the conformance counts that are written out by hand and are not interpolated at build time (`.zenodo.json`, `docs/getting-started.md`, the two landing-page meta descriptions and the two JSON-LD blocks on the conformance pages). `scripts/check_conformance_claims.py` now reports 450 checks, 53 domains and 294 standards, consistent across the corpus.

`git grep -nE '^(<{7} |>{7} )' -- .` is clean, so the new #432 gate is satisfied. On #428: the four `intensity_class` SVG variants regenerate byte-identical on the new base, so they are not stale, and `scripts/check_figure_contrast.py` passes over the whole set (1920 filled regions in 1272 figures clear dE00 10). The pass-region fill in this plot was already an opaque background-derived wash rather than an `alpha=` composite, which is why it clears the new threshold unchanged.

`ruff check .`, `mypy src scripts`, `bandit -r src`, `check_api_reference.py` (1014 names) and the full suite are green locally, and the fiche is still one page.

PR #422, opened 2026-07-29
Add the end-to-end HVAC duct-borne noise calculation and duct higher-order-mode cut-on

jmrplens:
Adds the end-to-end HVAC duct-borne noise calculation, the element models it
chains, and the higher-order-mode cut-on analysis that tells every plane-wave
method in the library where it stops being valid.

## The cascade

`duct_path` (new `noise_control/duct_path.py`) walks a fan-to-room path element
by element: subtract each attenuation, combine each element's regenerated sound
power on an energy basis over a 0 dB self-noise floor, apply the room effect,
rate the received spectrum against an NC or RC design curve.
`combine_duct_paths` adds a supply and a return path. `DuctPathResult` exposes
the calculation three ways:

- `.table()` returns the printed sheet row by row, with the four-row block per
  element the published worksheets use (attenuation with the worksheet's
  negative sign, *Sum*, *Self-noise*, *Combined*), the room effect, the received
  spectrum and the criterion curve;
- `.plot()` draws the cascade against the criterion curve;
- `.report()` renders a one-page sheet in the AHRI Standard 885 Table 8 layout.
  A path longer than a page elides its middle rows rather than spilling onto a
  second one; the complete sheet always stays available from `.table()`.

## The element models

New in `noise_control/hvac.py`, from Long, *Architectural Acoustics* 2nd ed.,
Chapters 13 and 14: `fan_sound_power` (Eq. 13.1 with the ASHRAE Tables
13.5-13.7) with `fan_efficiency_correction`, `blade_passing_frequency` and
`fan_casing_attenuation`; `unlined_rectangular_duct_attenuation` and
`unlined_circular_duct_attenuation`; `lined_rectangular_duct_attenuation` and
`lined_circular_duct_attenuation` (the Reynolds regressions);
`flexible_duct_insertion_loss`; `split_loss`; `diffuser_sound_power`
(Eqs. 13.27-13.33, the regenerated noise of a grille, register or diffuser
from its face area, approach velocity and pressure drop);
`end_reflection_loss_closed_form` and `equivalent_diameter`, also reachable as
`end_reflection_loss(method="long")` beside the existing Bies table look-up;
and `silencer_self_noise`. Plus `splitter_silencer_insertion_loss`, the
dissipative parallel-baffle unit reduced to lined airways and combined by Bies
Eq. (8.241); `room_effect`; and the two ASHRAE (2019) Applications Handbook
Chapter 49 screening tables, `air_terminal_velocity_limit` and
[...]

PR #423, opened 2026-07-29
Machine fault frequencies, experimental SEA and the plateau transmission-loss estimate

jmrplens:
Three additions anchored on Norton & Karczub, *Fundamentals of Noise and
Vibration Analysis for Engineers* (2nd ed., CUP 2003), each validated against
the book's own printed answers.

## Machine fault frequencies

New `vibration/machine_diagnostics.py` (Section 8.4). The kinematic families
every condition-monitoring job starts from:

- `bearing_fault_frequencies` - the seven rolling-contact bearing lines after
  Shahan and Kamperman (shaft, cage and its relative rate, rolling-element
  rotation and spin, and the outer- and inner-race element pass frequencies),
  Eqs. 8.4 to 8.14.
- `gear_mesh_frequencies` - mesh frequency, harmonics and modulation sidebands
  (Eq. 8.3).
- `induction_motor_frequencies` - supply, slip, pole-pass and rotor-slot
  harmonic lines with their eccentricity sidebands (Eqs. 8.19, 8.20).
- `blade_pass_frequencies` - blade tones of a fan, blower or pump plus the
  lobed interaction patterns of a ducted axial fan (Eqs. 8.15 to 8.18).

`FaultFrequencyResult` selects lines by name, in shaft orders or by frequency
span, merges families with `combine_fault_lines`, and its
`.plot(spectrum=...)` draws the predicted lines **over a measured envelope
spectrum**, which is the point of the feature: it closes the loop with the
`envelope_spectrum`, `cepstrum` and `time_synchronous_average` chain that was
already in `metrology` and had nothing to name its peaks with.

## Experimental SEA

New `vibration/experimental_sea.py` (Sections 6.3.3, 6.3.4, 6.4.1).
`power_injection_clf` inverts the two-subsystem steady-state power balance from
measured band energies (Eqs. 6.8, 6.10, 6.11 and 6.15), and
`power_injection_matrix` solves the full two-drive power-injection problem for
all four loss factors without assuming reciprocity, which then becomes a check
on the measurement rather than an input. The modal densities the inversion
needs come with it (Eqs. 6.23 to 6.29), including the three Szechenyi regimes
of a thin-walled cylinder about its ring frequency.

Two closed-form predictive counterparts join `junction_transmission` rather
than duplicating it: `right_angle_transmission_coefficient` (Eqs. 6.53 to 6.55,
which reproduces Norton Eq. 6.52 identically when fed to the existing
`coupling_loss_factor`, since that equation and Hopkins Eq. 2.154 are the same
expression once `cg = 2 cB`) and `point_connection_coupling_loss_factor`
(Eq. 6.56, bolts rather than a weld, falling as `1/f` instead of `1/sqrt(f)`).

## Plateau-method transmission loss

`plateau_transmission_loss` and `PLATEAU_MATERIALS` in
`building/panel_transmission.py` (Section 3.9.1, Table 3.1): the Watters
construction of a field-incidence mass law, a horizontal coincidence plateau
between points A and B set by the material's tabulated height and frequency
ratio, and a 10 dB per octave recovery above B.
`single_panel_transmission_loss` gains `coincidence_model="cremer"`
(Eq. 3.110) beside the existing Sharp model, and both it and
`mass_law_transmission_loss` accept an explicit `field_correction`.

## What each oracle validated

| Oracle (printed answer) | Validates | Agreement |
|---|---|---|
| Problem 8.5 (p. 619) | all seven bearing lines: `fs` 33,33; cage 13,80; element rotation 91,6; **BPFO 207**; spin 183,3; relative cage 19,53; **BPFI 293** Hz | exact at the printed precision |
| Problem 8.6 | blade-passing 350 Hz; lobed patterns 35 and 175 Hz | exact |
| Problem 8.7 | 60, 120, 360 and 3600 Hz | exact |
| Problem 6.10 | `eta12` **4,26e-4**, `eta21` **3,92e-4**, `Pi_in` **1,31 W** | within 0,4 % (see errata below) |
| Problem 6.13 | both `eta12` columns, welded and 12-bolt, 125 Hz to 2 kHz | within 0,7 % |
| Problem 3.11 | plateau TL 35,8 / 37 / 37 dB and the 10 dB/octave slope | exact where the construction is analytic; 0,8 dB below the chart-read values above point B |
| Problem 3.14 | field-incidence mass law 6,4 to 48,1 dB and Cremer 27 / 38,6 dB | within 0,1 dB |

Alongside the printed values the tests pin the exact identities:
`BPFO + BPFI = Z fs`, `BPFO = Z x FTF`, `BDF = 2 BSF`, `FTF + FTF_rel = fs`,
the Eq. (8.20) slot harmonic collapsing to `R fs` at any slip, SEA reciprocity
`n1 eta12 = n2 eta21`, the energy-balance round trip (input power equals total
dissipation), a forward-model round trip through both inversions, the plateau
degenerating to the mass law below point A, and the flat-plate modal density
agreeing exactly with the independent EN 12354-4 expression already in the
library.

## Errata found while anchoring

Four new entries in `docs/ERRATA.md`, each with a regression test pinning the
reading the library implements:

1. **Eq. (6.56)** is printed with the denominator bracket unsquared, which is
   dimensionally inconsistent. The squared form is what reproduces the book's
   own problem 6.13 answer.
2. **Problem 6.13 answer, `eta21` columns**: both are exactly the `eta12`
   column times `h2/h1`, i.e. the reciprocity step drops the plate area ratio
   `S1/S2 = 1,25` that Eq. (6.25) carries. Only the `eta12` columns are used as
   an oracle.
3. **Problem 6.10** states a 3,5 m x 3 m platform, but the three printed
   answers fix `E1/E2 = 6,554` while that geometry gives 7,88. The energy ratio
   is independent of the modal densities and of the wave speed, so only the
   plate area can reconcile it; the area the answers imply is 8,73 m2, which is
   3,5 x 3 minus the footprint of the cylinder standing on it in Fig. P6.10.
4. **Problem 3.14** states a loss factor of 1,5e-2, but its own answers above
   coincidence require 1,5e-3 (the discrepancy is exactly one decade of
   `10 lg eta` in Eq. 3.110).

## Docs and figures

- New guide **Machine fault frequencies** (EN + ES), pairing with the existing
  envelope-spectrum guide and walking a bearing-fault case end to end.
- New **Experimental SEA** section in the plate-junction guide (EN + ES).
- New **plateau method** section in the panel-insulation guide (EN + ES).
- Three figures through `scripts/generate_graphs.py` in all four variants: the
  fault-line overlay on a measured envelope spectrum, the predicted-versus-
  measured coupling loss factors, and the plateau estimate against the full
  physical model on the same panel.
- Both new modules registered in `scripts/api_taxonomy.py` and
  `docs/api-reference.md`; `docs/CONFORMANCE.md`, the API reference and the
  llms shards regenerated; CHANGELOG entry.

## Not anchored on an oracle

Two things are stated as such rather than presented as sourced:

- The **pole-pass frequency** `FP = p x f_slip` is standard
  condition-monitoring practice, not Norton's: he gives the slip frequency
  itself as the broken-bar sideband spacing and does not multiply it by the
  pole count. It is documented that way in the module and the guide, and only
  identity-checked in the tests.
- The **"non-diffuse field" column** of problem 3.14 (0, 3, 9, 15, 21, 27, 33,
  39 dB) is a constant 14,1 dB below the normal-incidence mass law in every
  band and I could not identify which relationship in the book produces it, so
  it is not implemented and not used as an oracle. The "diffuse field" column
  of the same table is fully reproduced.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added machine fault-frequency analysis for bearings, gears, motors, and fans, including spectrum overlays.
  * Added experimental SEA tools for modal-density calculation and coupling-loss-factor estimation from measured energies.
  * Added right-angle and point-connection junction analysis helpers.
  * Added plateau-based panel transmission-loss estimates and Cremer coincidence modeling.

* **Documentation**
  * Added comprehensive guides, API references, examples, figures, Spanish translations, and expanded Norton & Karczub errata coverage.


jmrplens:
Rerunning the automated review now that the earlier one hit the rate limit.

@coderabbitai review

jmrplens:
@coderabbitai full review

jmrplens:
@coderabbitai full review

jmrplens:
I went through this against the book itself (Norton & Karczub 2e, printed pp. 240-241, 397-399, 418, 520-521, 580-581, 593-594, 605, 611, 617, 619), re-derived the bearing kinematics and the SEA inversion from first principles, ran the envelope chain end to end on a synthetic defect, and mutation-tested the new constants. One blocking defect, a handful of non-blocking ones.

## Blocking

1. **`CHANGELOG.md` ships unresolved merge-conflict markers.** Lines 12, 68 and 136 of the branch file are `<<<<<<< HEAD`, `=======` and `>>>>>>> 9839e23f (Machine fault frequencies, ...)`. Both sides are Unreleased/Added content, so the resolution is to keep main's entries and append the three new ones, then delete the markers. Nothing in CI looks at the changelog, which is why it is green; worth a follow-up to add a conflict-marker grep to the gate.

## Non-blocking

2. **Five of the eight `PLATEAU_MATERIALS` rows are not pinned by any test.** `test_plateau_table_matches_norton_table_31` asserts brick, lead, steel and the key set only. I mutated aluminium's plateau height (29 to 32 dB) and glass's B/A ratio (10 to 6) and the entire suite still passes. Since the table *is* the feature, pin all eight tuples; I checked every row against the printed Table 3.1 (p. 241) and the transcription is correct as committed, so it is purely a regression-protection gap.

3. **The Cremer branch returns a float artifact at `f == fc`.** In `single_panel_transmission_loss(..., coincidence_model="cremer")`, the `np.maximum(f/fc - 1, tiny)` clamp turns Eq. (3.110)'s singularity into TL = -3051 dB: `single_panel_transmission_loss([2000.0], 15.0, critical_frequency=2000.0, loss_factor=0.01, coincidence_model="cremer")` gives -3051.4 with no warning. Small negative values just above `fc` are the formula's own behaviour and fine, but -3051 dB at the band that lands exactly on `fc` is neither Norton's number nor a usable one. Either exclude `f = fc` (raise or NaN with a note), floor it at the `10 lg eta`-limited dip, or document the open interval.

4. **Spanish decimal commas leaked into the English prose** (the rest of the EN pages use points, e.g. 12.96, 207.0):
   - `docs/machine-diagnostics.md:244` and `site/src/content/docs/guides/machine-diagnostics.mdx:256`: "crest factors of 2,5 to 3,5"
   - `docs/panel-sound-insulation.md:199` and `site/src/content/docs/guides/panel-sound-insulation.mdx:241`: "2,1 kg/m² per mm" and "35,8 dB"

   The same strings are baked into `llms-full.txt`, `llms-insulation-design.txt` and `llms-signals-spectra.txt`, so regenerate the shards after fixing.

5. **`_lobe_patterns` dedupes by lobe count alone and can drop a real pattern.** With N = 4 blades, V = 2 vanes and `harmonics=2`, the n = 2, k = 1 pattern (mL = 8 - 2 = 6, rotating at 2·4·fs/6 = 1.33 fs) is silently skipped because n = 1, k = +1 already produced a count-6 pattern at 0.67 fs. Patterns with equal lobe count from different blade harmonics rotate at different speeds and should coexist; include n in the name (`lobe n=2 m=6`) and dedupe on the pair, or on (count, frequency).

6. Two trivia, take or leave: the `target <= 0` guard in `plateau_transmission_loss` is unreachable (`plateau_height > 0` and `field_correction >= 0` force `target > 0`), and in the fan overlay the FTF/FTF_rel/shaft annotations overlap at low frequency on a wide axis.

## Verified correct

- **Bearing kinematics.** Problem 8.5's inputs are exactly Norton's (15 rollers, D = 34 mm, d = 6 mm, 12.96 degrees, 2000 r/min) and I re-derived all seven lines from the printed Eqs. (8.4) to (8.14) on pp. 520-521: FTF 13.80, FTF_rel 19.53, BSF 91.65 (the book's own 91.6 is rounded low, the test's asymmetric tolerance is the right call), BDF 183.3, BPFO 207.0, BPFI 293.0 Hz. The rpm-in/Hz-out boundary, the cos of the contact angle, the (8.7) vs (8.10) = 2x(8.7) split of BSF/BDF, and the invariance of BPFO/BPFI under `rotating_race` all match the book. `BPFO + BPFI = Z fs` is genuinely tested on three geometries, not just asserted in prose.
- **Envelope integration.** Synthesized an outer-race impulse train (shaft-rate modulated, ringing a 3 kHz resonance, buried in noise): `envelope_spectrum` peaks at 207.00 Hz against the predicted 207.01, and the `.plot(spectrum=...)` overlay places every dashed line at exactly the predicted frequencies. The shipped figure is honest about which lines the measurement does *not* hit.
- **Problem 6.10 and the area erratum.** The printed answers (4.26e-4, 3.92e-4, 1.31 W, p. 617) fix E1/E2 = (eta2 + eta21)/eta12 = 6.554, while the stated 3.5 m x 3 m platform gives 7.88, independent of modal density and wave speed; the 8.73 m² reading reproduces all three answers, and independently the printed eta21/eta12 = 0.920 equals n1/n2 only with the reduced area. The reciprocity test pins the printed pair against geometry-derived modal densities, which is a real check, not a recomputation; the two-drive inversion never assumes it, and its forward model in the test is written from Eqs. (6.10)/(6.11) directly.
- **Eq. (6.56) erratum.** Confirmed on the printed page 418: the denominator bracket is unsquared as printed and dimensionally inconsistent; the squared form gives 1.432e-2 at 125 Hz against the printed 1.44e-2 for the twelve-bolt pair (my independent computation), and reverting the code to the printed form fails the test, as it should. The eta21-column erratum of problem 6.13 is also real: every printed band is exactly eta12 x h2/h1 = 1.8333, dropping S1/S2 = 1.25.
- **Reuse.** The welded oracle goes through the existing `coupling_loss_factor` with cg = 2 cB (I recomputed 3.16e-3 at 125 Hz against the printed 3.15e-3), and `flat_plate_modal_density` is pinned exactly against the EN 12354-4 `modal_density` already in the library, so there are no conflicting conventions shipped.
- **Plateau method.** All eight Table 3.1 rows match the printed table; problem 3.11 recomputed by hand (35.8 on the mass law, 37/37 on the brick plateau, the chart-read 0.8 dB offset above B is what it is); problem 3.14 works with fc = 97.7/0.02 from Appendix 4 (not fitted) and requires the 1.5e-3 loss factor, confirming that erratum too. The estimate degenerates to the mass law below A at machine precision, is documented as an estimate throughout, and the comparison figure draws both curves on one panel with the same 5.5 dB field correction so the mass-law region overlays exactly.
- **Mutations that do get caught:** BSF's (1 - g²), the cage sign convention, Eq. (6.27)'s coefficient, the clf reciprocity ratio and the (6.56) bracket all fail tests when corrupted.
- **House rules:** no `assert` in `src/`, no em-dashes in any added line, no module-level level-2 relative imports in `_plot/`, all twenty new `__all__` names exported at the top level and present in `docs/api-reference.md`, EN and ES guides structurally parallel with translated prose and identical code, figures present in all four variants.

Fix the changelog and this is mergeable; 2 to 5 can ride along here or land as a follow-up.

jmrplens:
Thanks for the depth of this one, particularly for re-deriving the bearing lines from the printed equations rather than checking my arithmetic against itself. I went through all six findings; five are fixed, one I am pushing back on with a reproducer. The branch is now rebased on current `main` (#430, #432, #429, #428, #422 and #431 are all in) and every shared artefact was regenerated rather than hand-merged.

## 1. Conflict markers in `CHANGELOG.md` (blocking) - fixed

Confirmed exactly as reported, and worse than it looked: the markers were not only in the working tree, they were baked into the *content* of the branch's first commit, so replaying that commit on the new `main` produced nested markers instead of a clean two-way conflict. I flattened the branch to a single commit first, then rebased, which turned it into an ordinary conflict I could resolve properly.

The Unreleased/Added section now carries all nineteen entries: the five from this branch at the top (newest first, the convention #420 used) and the fourteen already on `main` below them, none dropped and none duplicated. I verified that mechanically rather than by eye, comparing the set of top-level bullets in the resolved file against the union of `origin/main` and the pre-rebase branch: nothing missing, nothing extra, and the whole-file diff against `main` is purely additive with zero deleted lines.

Two more errata conflicts appeared during the rebase (`docs/ERRATA.md` and the two Starlight pages, since `main` had grown the two Long entries meanwhile). Both sides kept, 44 sections on `main` plus the 4 new Norton ones = 48, and `scripts/generate_site_reports.py` re-run afterwards reports "4 pages checked, 0 written", so the transplanted copies match the source byte for byte.

Your gate passes locally:

```
$ git grep -nE '^(<{7} |>{7} )' -- .
$ echo $?
1
```

## 2. Five of eight `PLATEAU_MATERIALS` rows unpinned - fixed

Reproduced. The old test asserted brick, lead and steel plus the key set, so aluminium, concrete, glass, plaster and plywood were free. The table now goes in whole:

- `test_plateau_table_matches_norton_table_31` compares the entire mapping against a literal transcription of Table 3.1, so a corrupted number and an added material both fail;
- a parametrised `test_plateau_table_row_matches_the_printed_row` gives one case per material, so a failure names the row;
- a new parametrised `test_plateau_construction_uses_every_column_of_the_named_row` checks that the named route actually *reads* all three columns of its own row: the surface density through point A (inverting the field-incidence mass law at point A must return the plateau height), the plateau height as reported, and B/A placing point B. That closes the other half of the gap, where the table is right but the code reads the wrong row.

Mutation run over the restored file, five rows including your two:

```
aluminium plateau height 29->32   -> KILLED
glass B/A 10->6                   -> KILLED
plywood density 0.57->0.58        -> KILLED
plaster plateau height 30->31     -> KILLED
concrete B/A 4.5->4.6             -> KILLED
```

## 3. Cremer at `f == fc` - fixed

Reproduced: -3051.4 dB, and the `SoundReductionResult.transmission_coefficient` property turned that into a transmission coefficient of order 1e305, which is the part that really bothered me.

Of the three options you offered I took the floor, because the book supplies the value. Two paragraphs above Eq. (3.110), on the same page as Eq. (3.109): "At the critical frequency, theta = 90 degrees, and the panel offers no resistance to incident sound waves." That is tau = 1, TL = 0 dB, and it is also the hard bound of any passive panel, so the floor is not a fudge on the empirical line but the physical limit the empirical line is approximating. The band on `fc` now reports 0.0 dB and no band can report tau > 1. I preferred it to raising, which would have been arbitrary: rejecting exactly `fc` while happily returning -120 dB at `fc * (1 + 1e-9)` is not a defensible boundary.

Nothing that was verified moves: the problem 3.14 column is untouched (its printed answers are all positive and `fc` = 4885 Hz is not a band centre), and so is the 10 dB per octave slope far above coincidence. `test_cremer_is_floored_at_the_coincidence_frequency` pins the new behaviour and the docstring states the interval and its source.

## 4. Spanish decimal commas in the English guides - fixed

All four spots corrected to points, in `docs/` and in the Starlight pages, and `llms-full.txt`, `llms-insulation-design.txt` and `llms-signals-spectra.txt` regenerated from source. Grepping the shards for the old strings now returns nothing. I left the commas in `docs/ERRATA.md` alone deliberately: that file quotes printed values from standards and books that use the comma themselves, and the 111 pre-existing occurrences on `main` set the convention.

## 5. `_lobe_patterns` dropping distinct patterns - fixed

Confirmed against Eqs. (8.16) and (8.17) on the printed page: the pattern speed is `n N (r/min) / mL`, with `n` in the numerator, so the lobe count alone does not identify a pattern. Your N = 4, V = 2, harmonics = 2 case reproduces exactly, and after the fix:

```
lobe n=1 m=2    2.000 fs
lobe n=1 m=6    0.667 fs
lobe n=2 m=6    1.333 fs     <- was silently dropped
lobe n=2 m=10   0.800 fs
```

De-duplication is now on the `(n, mL)` pair, which still collapses the genuine duplicates (within one `n`, `+k` and `-k` can reach the same `|mL|` at the same speed, for instance N = 2, V = 4). The lines are named `lobe n=1 m=2` as you suggested, so the printed problem 8.6 answers keep their oracle, and the docstring, both guides and the regenerated API pages carry the new names with a sentence explaining why `n` is in there. A new test pins both the two m = 6 speeds and the full ordered list of the four patterns.

## 6a. The `target <= 0` guard - pushing back, it is reachable

This one I disagree with, and I would rather cover it than delete it. `plateau_height > 0` and `field_correction >= 0` do not force `target > 0`, because the expression underflows before the algebra does: `10.0 ** ((height + correction) / 10.0)` rounds to exactly 1.0 for a subnormal height with no field correction, so `target` is exactly 0.0 and point A would land at 0 Hz, putting the entire spectrum on the plateau.

```python
>>> plateau_transmission_loss([100.0, 1000.0], mass_per_area=20.0,
...                           plateau_height=1e-300, frequency_ratio=4.5,
...                           field_correction=0.0)
ValueError: the plateau height sits below the mass law at every frequency; ...
```

The guard stays, and `test_plateau_rejects_a_height_that_underflows_point_a` now covers it, so it no longer reads as dead code to the next person through here.

## 6b. Overlapping annotations - fixed

Reproduced on a bearing result drawn on a 1200 Hz axis: FTF (13.8 Hz), FTF_rel (19.5 Hz) and shaft (33.3 Hz) collided into an unreadable block, and BDF/BPFO touched. Since the names are drawn rotated, each occupies a narrow vertical strip, so a vertical stagger would not have helped and the fix has to be horizontal.

The labels are now laid out across the axis in points rather than in hertz: walking the lines in frequency order, each label is pushed just far enough right to clear the previous one, which is the least displacement that separates a group. An isolated line keeps its label exactly where it was, and the drift through a dense run is bounded by the difference between the label width and the line spacing rather than growing with a fixed step. My first attempt was a fixed ladder and it failed its own test case: with 25 Hz gear sidebands the 8 pt step was *wider* than the spacing, so a stepped label landed on top of the next unstepped one. The seven-sideband `2xGMF` family is readable now, which it was not before this branch either.

The shipped `bearing_fault_envelope` figure regenerates byte-identical in all four variants, because none of its lines is crowded at that scale, and `test_crowded_fault_labels_do_not_overlap` pins both the clearance and the "isolated label does not move" half.

## Picked up from the rebase

The figure contrast gate that landed with #428 caught two fills of mine straight away, so both are now on the new convention:

```
- experimental_sea_clf.svg:        patch_10 fill #e0e0e0 @ 1.00 on #ffffff  dE00 6.50
- plateau_transmission_loss.svg:   patch_3  fill #d62728 @ 0.12 on #ffffff  dE00 9.86
- plateau_transmission_loss_dark:  patch_3  fill #d62728 @ 0.12 on #000000  dE00 9.34
```

The SEA bar chart drew the `eta_2` bar in `COLOR_GRID`, which is exactly the colour tuned to vanish into the page, so it is `COLOR_MUTED` now. The coincidence plateau was a fixed `alpha=0.12` wash, so it goes through `theme_fill` in both the library renderer and the figure script. The plateau band now reads as a pale tint on the white page and a raised shade on the dark one.

One more that is not mine but was blocking the job: `duct_mode_cut_on`, from #422, still draws its plane-wave band at a fixed `alpha=0.10` and fails all four variants (dE00 5.27 light, 4.23 dark). It came in after the gate, so `main` carries it. It is a one-line change of exactly the kind #428 established, so it rides here as its own commit rather than leaving the figures job red on every branch that rebases. Happy to pull it out if you would rather it landed separately. With it, the whole set reports "All 1938 filled regions in 1288 figures clear dE00 10".

## Gates

`ruff check .`, `mypy src scripts`, `bandit -r src`, the full `pytest` run including `tests/test_package_architecture.py`, `scripts/check_api_reference.py`, `scripts/check_figure_contrast.py`, the regenerated `docs/CONFORMANCE.md` (re-run afterwards: no drift), API reference, `llms` shards and site reports are all green locally on the rebased branch.

jmrplens:
Retracting the `duct_mode_cut_on` commit I mentioned above: #434 already owns that fix, so mine was a duplicate and is now dropped from this branch. The branch head is `1c803d01` and touches nothing outside its own scope again. I will rebase onto `main` once #434 lands so the figures job picks the fix up from there.

PR #424, opened 2026-07-29
Add a security policy and a pull request template

jmrplens:
Two community health files were missing: a security policy and a pull request template. This adds both, written from what the project actually is rather than from a template.

## `SECURITY.md`

A generic policy would be worse than none here, so the threat model is specific to a computation library that runs no service and holds no credentials:

- **Supported versions.** Fixes ship forward on the latest 3.x with no backports, on the interpreters CI actually exercises (3.13 and 3.14, Ubuntu/macOS/Windows). The `PyOctaveBand` stub is called out as unmaintained.
- **Reporting.** Private reporting through GitHub security advisories, with the maintainer address already published in `CITATION.cff` and `CODE_OF_CONDUCT.md` as the fallback. Acknowledgement in 5 working days, first assessment in 15, fix with the next release. Deliberately not a 24 hour SLA for a single-maintainer research library.
- **Coordinated disclosure**, with 90 days from acknowledgement as the ceiling and credit in the advisory and the changelog.
- **Scope.** In: untrusted input reaching a reader (the ANP CSV loader is the concrete case), deserialisation, the release supply chain, repository automation. Out: dependency CVEs, resource exhaustion you asked for, out-of-range inputs, scanner output with no impact, and most importantly a value that disagrees with a standard. That last one is a conformance defect, and the policy links the conformance defect form and the Standards & conformance discussion category for it. Defects in the published standard itself are pointed at `docs/ERRATA.md`, explicitly not security advisories.
- **What already runs**, described rather than invented: CodeQL default setup, bandit in the `quality` job (any finding fails, including Low), GitGuardian on every pull request, Dependabot for pip/npm/Actions, the `snyk` Makefile target, and the conformance staleness gate as an integrity control on the numbers themselves.

## `.github/pull_request_template.md`

[...]

PR #426, opened 2026-07-29
Weston propagation regimes, marine-mammal exposure criteria and the detection-range solver

jmrplens:
Four additions to the underwater domain, all anchored on published numbers.

## Weston's shallow-water propagation regimes

New `underwater/weston_regimes.py`. A source in shallow water loses energy in
four successive range regimes, and the boundaries between them follow from the
seabed reflectivity alone:

| Regime | Propagation factor | Loss law | Ends at |
|---|---|---|---|
| Spherical | 1/r² | 20 lg r | H/(2·ψc) |
| Cylindrical | 2·ψc/(rH) | 10 lg r | r_CS = π·H/(4·η·ψc²) |
| Mode stripping | √(π/(η·H))·r^−3/2 | 15 lg r | r_MS = k²·He³/(9·π·η) |
| Single mode | exponential | steeper | — |

`weston_propagation_loss` assembles the composite loss and returns each
regime's own law beside it; `weston_regime_boundaries` gives the three
transition ranges, the Weston effective depth, the waveguide cut-off frequency
and the number of cut-on modes. `reflection_loss_gradient`,
`critical_grazing_angle`, `effective_depth`, `waveguide_cutoff_frequency` and
`loss_parameter` expose the underlying closed forms, and the two characteristic
seabeds ship as `WESTON_SEABEDS`.

Beyond being a feature, this is the independent verification anchor the
numerical solvers were missing. The energy-flux result is incoherent, so it
describes the range-averaged field: set the critical angle to 90° and the loss
gradient to zero and the cylindrical branch reduces to F = π/(rH), the exact
many-mode limit of an ideal waveguide. The depth- and range-averaged
`normal_modes` loss lands on it to within 0.55 dB at 50, 100 and 200 Hz, and
`parabolic_equation` to within its known paraxial bias. Both are wired as
tests, and the modal one also as a conformance check.

## Marine-mammal audiograms and regulatory auditory weighting

New `underwater/marine_mammal_audiograms.py` and
`underwater/marine_mammal_weighting.py`.

- `group_audiogram` implements the Southall et al. (2019) group audiogram
  (their Equation 1) in both the absolute Table 2 fit and the normalised
  Table 3 one. Low-frequency cetaceans are deliberately absent: the article
  publishes no fitted audiogram for them and never prints F1.
- `orca_audiogram` implements the three-branch killer-whale audiogram of
  Ainslie Eq. (11.159). The published 51.2 dB at 50 kHz needs the third
  branch; the second one returns 50.5 dB there, and both points are pinned.
- `auditory_weighting` implements the band-pass weighting and exposure
  functions **with the guidance version explicit and carried on the result**:
  NMFS 2024 v3.0 by default because it is current, with NMFS 2018 v2.0 and
  Southall et al. 2019 selectable, since assessments in flight still cite the
  older ones. Group codes are not portable between versions (NMFS 2018 "HF" is
  Southall "VHF"), so passing the wrong code raises instead of silently
  returning the wrong filter.
- `exposure_criteria` returns the published TTS and injury (PTS / AUD INJ)
  onset criteria, including the unweighted peak-SPL half of the impulsive dual
  metric.
- `weighted_exposure` runs the assessment end to end: weighting a band
  spectrum, accumulating it over events (the ISO 18406 +10·lg N) and reporting
  the margin against every applicable criterion.

The pile-driving module is wired into that chain: `strike_sel_spectrum` splits
the single-strike sound exposure of a recorded strike into fractional-octave
bands by Parseval, so a piling campaign now goes from waveform to verdict.

## Detection range

`detection_range` and `detection_range_from_curve` in
`underwater/sonar_equation.py`. The figure of merit is the maximum allowable
transmission loss, so inverting a loss law at TL = FOM gives the range at which
the detection probability is 50 %. The first inverts the closed-form loss
(monotone, single root); the second reads the crossing off any computed curve,
including the oscillatory loss of a real waveguide, where `crossing` selects
the first or the last one.

## Medwin sound speed

`sea_water_sound_speed(..., model="medwin")`, the simplest member of the
family, alongside UNESCO/Chen-Millero, Del Grosso and Mackenzie.

## Oracles

Every number in the tests comes from a printed source, not from this code:

- **NMFS 2018 Appendix D worked example**: W(1 kHz) for the five hearing
  groups (−0.06, −29.11, −37.55, −5.90, −4.87 dB), reproduced to ±0.01 dB.
- **C = −max W(f)** recomputed for all three parameter sets (5 + 7 + 8 rows),
  and **Tw = K + C** against the published weighted TTS onset column.
- **Injury = TTS + 20 dB** (non-impulsive) for every group of every version,
  and Southall's impulsive offsets (−11 dB SEL, +15 dB PTS SEL, +6 dB PTS
  peak, threshold at f0 + 159 dB).
- **Southall's three prose audiogram thresholds** at f0 (54, 48 and 53 dB
  re 1 µPa) from Equation (1) plus Tables 2 and 4.
- **Ainslie's orca audiogram** at both published points, 39.0 dB at 22.6 kHz
  and 51.2 dB at 50 kHz, with the wrong-branch value pinned as a regression.
- **Ainslie Table 9.1** seabed properties: ε = 0.0161 / 0.00165, η = 0.28
  Np/rad for sand and 0.021·f̂ for mud.
- **Ainslie's seven numeric sonar worked examples**: every published figure of
  merit (78.0, 79.0, 82.7, 73.8, 69.8, 59.0 and 51.0 dB re m²) and the
  detection ranges 1.3 km, 0.9 km and 241 / 149 / 117 m.
- **Medwin's two published partial derivatives** (∂c/∂T = 4.6 − 0.110·T, 3.5
  m/s per °C at 10 °C; ∂c/∂z = 0.016 m/s per m) and cross-model agreement.

Eight new conformance checks (437 total).

## Errata

Three new entries in `docs/ERRATA.md`:

1. **Ainslie Eq. (9.57)** prints r_MS = k²·He³/(9·η), a factor π larger than
   the value its own stated derivation rule gives (equating θ_eff with θ_n at
   n = 3/2). Verified symbolically and numerically; the library implements the
   consistent form and pins the equating rule itself as the test.
2. **NMFS 2024 Table 5** prints C = 1.37 dB for otariids in water where 1.36
   is correct. NMFS says so itself in the table footnote, and recomputing C as
   the peak of W gives 1.3643. The library implements 1.36 and keeps the
   printed digit reachable as `WeightingParameters.c_db_as_printed`.
3. **Southall et al. Table 7** carries four peak-SPL misprints, corrected by
   the journal's own errata (Aquatic Mammals 45(5)). The corrected values are
   the ones implemented.

The registry's stated scope covers "standards, guidance documents and
technical reports, textbooks and journal papers", so all three fit; the two
non-standard sources are marked as such in their entries.

## Docs

New bilingual guide "Marine-mammal noise exposure" (weighting, criteria and a
worked pile-driving case), new "Weston's shallow-water regimes" and "Detection
range" sections of the underwater propagation guide, two new figures
(`weston_regimes`, `marine_mammal_weighting`) in all four variants, and the
usual bibliography, section-overview, sidebar, API-reference, conformance and
CHANGELOG updates.

## Not anchored

The two passive worked examples of Ainslie chapter 3 quote detection ranges of
2.5 km and 10.2 km from a coherent Lloyd-mirror propagation model that the
closed-form spreading laws here do not reproduce, so only their figures of
merit are tested, not their ranges. The Southall LF-cetacean audiogram is not
implemented at all, because F1 is never printed.

## Summary by Sourcery

Add analytic Weston shallow-water propagation regimes, marine-mammal audiograms and regulatory auditory-weighting/exposure tools, a detection-range solver for the sonar equation, a Medwin sea-water sound-speed model, and corresponding plotting, docs, tests, errata and API wiring.

New Features:
- Introduce Weston shallow-water propagation regime modelling, including regime boundaries, seabed parameters and a composite propagation-loss API.
- Add marine-mammal audiogram support and regulatory auditory weighting, exposure criteria and end-to-end weighted-exposure assessment utilities across NMFS and Southall guidance versions.
- Provide detection-range computation from closed-form transmission-loss laws and arbitrary loss curves within the sonar-equation module.
- Implement the Medwin (1975) empirical sea-water sound-speed formulation alongside existing UNESCO, Del Grosso and Mackenzie models.
- Expose pile-driving single-strike SEL spectra resolved into fractional-octave bands for use in marine-mammal exposure assessments.

Bug Fixes:
- Document and compensate for published numeric errata in Ainslie Eq. (9.57), NMFS 2024 otariid weighting parameter C and Southall et al. Table 7 peak-SPL values, ensuring consistent internal implementations.

Enhancements:
- Extend underwater plotting with new renderers for Weston regimes, marine-mammal audiograms, auditory weighting functions, weighted exposure, strike SEL spectra and detection range, including Spanish i18n.
- Expand documentation and guides with new sections on Weston regimes, detection range and marine-mammal noise exposure, plus new reference figures and bibliography entries.
- Wire new underwater functionality into the public API namespaces and update the API taxonomy and summaries to reflect expanded underwater and fauna capabilities.

Documentation:
- Update conformance report counts and errata registry to reflect new underwater and marine-mammal validations and documented source issues.
- Revise underwater propagation guides (EN/ES) and section overviews to include Weston regimes, Medwin sound speed, detection range and marine-mammal exposure, and adjust README and API-reference descriptions accordingly.

Tests:
- Add extensive conformance checks for Medwin sound-speed derivatives, Weston regime seabed gradients and flux-versus-modal averages, NMFS and Southall auditory-weighting parameters, orca audiogram values, sonar worked examples and detection-range behaviour, increasing the numerical conformance coverage.
- Introduce unit tests for pile-driving band SEL spectra, weighted exposure chaining and underwater plotting localisation of new renderers.

Chores:
- Refresh generated graphs and i18n mappings to include new Weston and marine-mammal weighting figures in both English and Spanish variants.

## Summary by CodeRabbit

* **New Features**
  * Added Weston shallow-water propagation regimes and seabed modeling.
  * Added sonar detection-range calculations, including curve-based inversion.
  * Added Medwin sea-water sound-speed support.
  * Added marine-mammal audiograms, auditory weighting, and exposure assessments.
  * Added strike-level, frequency-band sound exposure calculations and plotting.
* **Documentation**
  * Added guides, API references, examples, citations, and errata for the new underwater capabilities.
* **Quality**
  * Expanded validation and conformance coverage to 467 passing checks across 55 domains and 311 standards.


jmrplens:
@coderabbitai review

jmrplens:
## Review

This one got no automated review at all (CodeRabbit hit the plan rate limit, Sourcery declined on diff size, Bito on its trial limit), so I went through the sources myself: Ainslie §9.1.1.2 and §11.4.6, NMFS 2018 Appendix D, NMFS 2024 Table 5 / Table ES2 / Table A.E-2, and Southall et al. (2019) including the errata bound into the PDF.

Most of it holds up well, and the oracle discipline is real: I sampled well beyond eight test groups and every published number I checked is what the source prints, fed with the source's own inputs. Two blocking items and a handful of smaller ones below.

---

## Blocking

### 1. `detection_range_from_curve` returns `inf` when the target is undetectable everywhere

`src/phonometry/underwater/sonar_equation.py`, the `up.size == 0` branch.

```python
below = tl <= fom
up = np.flatnonzero(below[:-1] & ~below[1:])
if up.size == 0:
    return float("inf")
```

If the loss already exceeds the figure of merit at the first sample, there is no upward crossing and the function returns `inf`. But `inf` is the value that means "the loss never reaches the FOM", i.e. detectable at any range. The two opposite physical situations collapse onto the same return value, and the one that is returned is the optimistic one:

```
detection_range_from_curve(20, [1,2,3], [50,58,70])  ->  inf
```

The loss is 30 dB above the FOM at every range; the answer should be 0, not "unlimited". A purely descending curve does the same. The closed-form sibling `detection_range` gets this right and returns `0.0` when `_loss(lo) > fom`, so the two functions disagree on the same question. Please make the curve version agree, and say in the docstring which end `inf` belongs to.

### 2. The Ainslie Eq. (9.57) errata entry does not describe what the library implements

The headline claim is correct and I am happy to see it recorded: the print really is a factor pi too large relative to its own stated derivation rule. But the entry as written mixes two different derivations, so a reader who follows it cannot reproduce either the formula or the numbers.

Eq. (9.47) as printed is `theta_eff = (pi*H/(4*eta*r))^0.5` with **H**, the true water depth. Eq. (9.56) is `theta_n = n*pi/(k*He)` with **He**. Equating them at n = 3/2 gives

```
r = k^2 * He^2 * H / (9*pi*eta)
```

The errata entry instead writes the equating step as `pi*He/(4*eta*r) = ...`, silently substituting He for H, and concludes `r = k^2*He^3/(9*pi*eta)`. That is also what `weston_regime_boundaries` implements. Both are defensible, but the entry then quotes evidence belonging to the *other* one. For H = 50 m, f = 250 Hz, c = 1500 m/s, eta = 0.28 (He = 53.628 m):

| form | value |
|---|---|
| A. strict Eq. (9.47), uses H: `k²·He²·H/(9·π·η)` | 19.92 km |
| B. He-substituted: `k²·He³/(9·π·η)` (what the library computes) | 21.36 km |
| C. printed Eq. (9.57): `k²·He³/(9·η)` | 67.12 km |

The entry quotes "19,9 km" and "67,1 km", which is A against C, and says "the ratio is exactly pi·He/H", which is again A against C (3.3695). But it claims the derivation gives B, whose ratio to C is exactly pi (3.1416). So the entry's own evidence line contradicts its own conclusion, and the 19.9 km it quotes is not what the shipped code produces (21.4 km).

Pick one reading and make the entry, the code and the numbers agree. If you keep B, say explicitly that Ainslie also carries the He-for-H substitution into theta_eff, and quote 21.4 km with a ratio of exactly pi. If you keep A, change the implementation.

Related, and part of the same fix: `test_mode_stripping_boundary_equates_theta_eff_with_mode_3_over_2` (tests/underwater/test_weston_regimes.py:119) says in its docstring that it pins "the equating rule itself", but line 131 writes

```python
theta_eff = np.sqrt(np.pi * b.effective_depth / (4.0 * b.reflection_loss_gradient * r))
```

with `effective_depth` where Eq. (9.47) has H. So it does not exercise the published rule, and both of its assertions are restatements of the implementation formula. Note the module's own `weston_propagation_loss` uses `h` here (line 502), correctly, so the library is currently inconsistent with itself between the composite loss and the boundary.

For what it is worth, an independent physical check supports the corrected value rather than the print: at 21.5 km the mode-stripping factor (Eq. 9.49) and the peak single-mode factor (Eq. 9.54 with the sin² terms at maximum) cross within about 15 %, whereas at 67.6 km the single-mode branch has already overtaken mode stripping by more than a factor of 1.5. Worth adding to the entry, since it is evidence that does not depend on how you read the derivation.

---

## Non-blocking

### 3. `strike_sel_spectrum` emits NaN bands that `weighted_exposure` rejects

Empty bands get `np.nan` (`np.where(band_energy > 0.0, band_energy, np.nan)`), and `weighted_exposure` raises `ValueError: 'band_sel' must be finite`. Any 1/3-octave band narrower than the FFT bin spacing is empty, which is the normal case for a single strike:

```
1.0 s @ 48 kHz  ->  0/34 NaN bands, chain OK
0.2 s @ 48 kHz  ->  1/34 NaN bands, ValueError
0.1 s @ 48 kHz  ->  3/34 NaN bands, ValueError
0.5 s @ 10 kHz  ->  6/34 NaN bands, ValueError
```

The guide works around it with `finite = np.isfinite(spectrum.band_sel)` and masks, which is energy-neutral and correct, but neither docstring mentions the sentinel and the PR description says the chain goes "from waveform to verdict". The obvious call raises. Either document the NaN convention on `StrikeSelSpectrum.band_sel` and on `weighted_exposure`, or return `-inf`/drop empty bands so the advertised chain works unaided.

### 4. The Southall Table 7 entry's "Evidence" paragraph misreads the source

The four corrected values are exactly right, verbatim from the errata (I checked: PCA 138 -> 155 and 144 -> 161, OCA 161 -> 170 and 167 -> 176). The justification underneath is not.

- "the corrected values satisfy the article's own derivation rules where the printed ones do not: PTS peak = TTS peak + 6 dB". Both satisfy it. 144 − 138 = 6 just as 161 − 155 = 6, and the entry concedes this in the same sentence. It discriminates nothing.
- "the 15 dB nominal offset between the SEL-based and peak-SPL-based in-air thresholds stated on printed p. 155". That is not what p. 155 says. The 15 dB is the offset from SEL TTS onset to SEL PTS onset ("this approach prescribes adding 15 dB to the TTS-onset threshold to estimate PTS onset"), not an SEL-to-peak offset, and it is not in-air specific.

The genuinely discriminating rule is on the same page: peak SPL TTS onset for groups without data is the threshold at f0 plus 159 dB. Running our own audiogram gives PCA T(f0) = −4.56 and OCA T(f0) = 11.41 dB re 20 uPa, so the rule predicts 154.4 and 170.4, matching the corrected 155 and 170 and nothing like the printed 138 and 161. That is real independent evidence and it belongs in the entry in place of the current paragraph.

### 5. Extend the +159 dB test to PCA and OCA

`test_southall_impulsive_peak_spl_is_threshold_at_f0_plus_159_db` only exercises PCW. Applying it to PCA and OCA would turn the Table 7 errata claim from a transcription into a derived result, which is exactly the anchor that entry wants. (Minor: the PCW assertion currently passes at the exact tolerance boundary, `round(53.501) + 159 = 213` against `approx(212, abs=1.0)`.)

### 6. Mutation-coverage gaps in the published tables

Corrupting these rows would not fail anything:

- `hearing_range_hz` on every row of both NMFS tables (24 published numbers from Table ES1) is never referenced by any test.
- `_CRITERIA_2024_IMPULSIVE` is pinned only at VHF (196/202). Table A.E-2 satisfies injury SEL = TTS SEL + 15 and injury peak = TTS peak + 6 on all eight rows, but `test_southall_impulsive_offsets` loops over `hearing_groups("southall-2019")` only. Widening it to all three guidance versions costs one line and covers 28 numbers.
- `_CRITERIA_*_CONTINUOUS.tts_sel` is never tied to `K + C`, even though NMFS's own footnote says they are "mathematically equivalent to K + C". The test uses a separate hard-coded `_PRINTED_TTS` dict, so the two copies of the same published numbers can drift apart silently. Assert `exposure_criteria(g).tts_sel == round(K + C)` instead.

### 7. Pin the PE bias to a number

"Within its known paraxial bias" is `abs=3.0`. The actual deviation is **+2.05 dB** (numeric 61.00, flux 58.95). Because the band is symmetric around the flux value while the truth sits at +2.05, the test currently accepts anything from −5.05 to +0.95 dB of error, and in particular a PE that reproduced the flux law exactly would pass. Assert a band, e.g. `1.5 < deviation < 2.6`. The modal check is fine: −0.29 / −0.55 / −0.15 dB against `abs=1.0`, and the PR description's "within 0.55 dB" is accurate.

### 8. `weighted_exposure` results alias the caller's arrays

`res.band_sel is band_sel` is `True`, likewise for `frequencies`. The dataclass is `frozen=True`, which reads as immutable, but mutating the caller's array afterwards silently changes an already-returned result and vice versa. A `np.array(..., copy=True)` on the way in would settle it.

### 9. `_any_positive` treats a margin of exactly zero as an exceedance

`any(m is not None and m >= 0.0 ...)`. At exactly the criterion the result reports the criterion as exceeded. Defensible as precaution, but it should be stated in the `exceeds_injury` docstring, since the natural reading of "exceeds" is strict.

### 10. Minor: the v = 2 m/s detection-range anchor

At v = 2 the wind term is `22.4*lg(1) = 0`, so that case carries no information about the scaling and is purely a check of the closed-form propagation against Ainslie's. It lands at 234.3 m against a printed 241 m, i.e. −2.8 % with only 2.2 points of headroom in the 5 % band. Fine as it stands, but it is the one that will go red first if the absorption default ever changes.

---

## Verified as correct

- **NMFS 2024 otariid C.** The footnote is verbatim as quoted ("an error was identified with the Navy's rounding, where this value should be 1.36, instead of 1.37"). Recomputing C as the negated peak of W from a = 1.58, b = 5, f1 = 2.53, f2 = 43.8 gives **1.3643**, and I recomputed all seven rows: every other one rounds to its printed value, OW is the only disagreement. The entry is sound and the `c_db_as_printed` escape hatch is the right call.
- **Southall Table 7 values.** All four match the errata bound into the PDF exactly, and the Table 5 "B" vs "b" note is there too.
- **Ainslie Eq. (9.57), core claim.** The print is `k²·He³/(9·η)` and the text does say "estimated by equating theta_n and theta_eff with n = 3/2". The factor pi is real under either reading of theta_eff. Only the entry's presentation needs fixing, not its conclusion.
- **The removed float-equality guard is genuinely unreachable.** I checked every path. `_finite_array` rejects NaN and inf up front; equal consecutive samples cannot appear in `up` because `below[i] == below[i+1]` there; `crossing="last"` draws from the same set; descending and non-monotonic curves are handled pointwise by the mask; and `tl[i] <= fom < tl[i+1]` in IEEE 754 guarantees a strictly positive difference, with `frac` confined to [0, 1). Removing it was correct. (The defect above is three lines up, not here.)
- **NMFS 2024 Table A.E-2.** I rendered the page and checked all eight rows against `_CRITERIA_2024_*`. Every one of the 48 numbers matches, including the OCW/PCW/OCA/PCA to OW/PW/OA/PA renaming. Leaving SI out for nmfs-2024 is right, since Table ES2 publishes no weighting parameters for it.
- **NMFS 2018 Appendix D.** The five W(1 kHz) values are printed exactly as the test claims, and the hand-evaluated LF row is a genuine independent term-by-term check.
- **Orca audiogram.** All three branches, both break points and both check values match Eq. (11.159) and its surrounding prose (39.0 dB at 22.6 kHz, 51.2 dB at 50 kHz). The third-branch test is honest, not circular: it computes the second branch independently of the API, so moving the break would fail both the 51.2 assertion and the 0.5 dB separation. Good regression.
- **The LF-cetacean omission.** Confirmed: Tables 2, 3 and 4 have no LF row, the prose says Eq. (1) was fit "for each marine mammal group except LF cetaceans", and while A = 20 dB/decade, B = 3.2, F2 = 9.4 kHz and T0 = 53.2 dB are all given in prose, F1 is only described by the criterion used to choose it. The error message is accurate and not covering a gap.
- **Eq. (9.60).** `(np.pi - bed.density_ratio)` looks like a transcription slip but is not: the print really is `(pi − rho_sed/rho_w)`, and it follows from Eq. (9.58) with Eq. (9.55) substituted. Faithful.
- **Table 9.1.** eps = beta/(40·pi·lg e) reproduces 0.0161 and 0.00165 (the plain 40·lg e would give 0.0507), eta_sand comes out 0.2785 against a printed 0.28, and eta_mud is 0.0207·f against 0.021·f.
- **The wind-noise scaling.** Eq. (11.165) is `10^-7.12 · v^2.24`, the Figure 11.22 caption prints "117, 149, and 241 m, respectively, for v10 = 10, 6, and 2 m/s", and Ainslie states outright that wind is the only noise source, which is what licenses scaling all of NL. The 5 % tolerance is **not** too loose here: I swept the exponent and only 2.24 passes; 2.0 and 2.5 both fail, as does swapping the spreading law.
- **The untested Chapter 3 passive ranges.** Justified. Ainslie's Eq. (3.86) is a coherent Lloyd-mirror result with an r⁻⁴ dependence at 300 Hz; the closed form gives 7.9 km against the printed 2.5 km. That is a model difference, not a broken spreading law, and `test_detection_range_inverts_spherical_spreading_exactly` pins the closed form independently.
- **Parseval split.** The one-sided weighting (1, 2, ..., 2, 1 for even n) and the `1/(fs·n)` normalisation are right; a full-band check reproduces the direct time-domain energy, and the shortfall against `broadband_sel` is exactly the energy above the top band edge, as documented.
- **`guidance=` is not decorative.** W(1 kHz) for LF is −0.0303 under nmfs-2024 against −0.0644 under nmfs-2018/southall-2019, the peak criteria differ (222 vs 219), and the injury label switches between AUD INJ and PTS.
- **House rules.** No bare `assert` in `src/`. No module-level level-2 relative imports in `_plot/underwater.py`. All 82 new `__all__` names appear in `docs/api-reference.md`. EN/ES parity is clean on all three guide/reference pairs: identical numeric multisets, identical heading counts and identical code-block counts. The 15 added em-dashes are all either ISO standard titles (established style, 101 of them already in `references.md` on main) or table cell placeholders, not prose.

jmrplens:
Thank you, this was a genuinely useful review, and given that no bot managed to look at the diff it was the only scrutiny this branch got. I verified every finding against the sources before touching anything. Nine of the ten are accepted and fixed; on one I reached a different conclusion from yours, and two of your supporting arguments do not reproduce. Point by point below.

The branch is rebased onto current `main` (#420, #422, #424, #427, #428, #429, #430, #431, #432) with the conformance report, api-docs, llms and figures regenerated rather than hand-merged.

---

## Blocking 1: `detection_range_from_curve` returning `inf` for the undetectable case

Accepted in full, and thank you for catching it. The two situations really did collapse onto the optimistic value, and the two solvers really did disagree.

The rule is now decided by the loss at the **last** sample. If there is no upward crossing and the loss is still below the figure of merit there, the target stays detectable past the end of the grid and the answer is `inf`. If it is above, then since reaching "above" from "below" would itself have produced an upward crossing, it was above at every sample, and the answer is `0.0`, which is what `detection_range` returns for the same case. That also settles the descending-curve case you mentioned: a curve falling through the figure of merit ends below it, so it correctly returns `inf`.

The docstring now says which end `inf` belongs to, and `DetectionRangeResult.detection_range` documents both limits.

Pinned by `test_both_solvers_return_zero_when_the_target_is_undetectable_everywhere`, which does not just assert `0.0` twice. It takes the loss curve that the closed-form solver returns for a figure of merit it cannot reach, checks that the minimum of that curve really is above the figure of merit, and feeds the curve back into `detection_range_from_curve` with the same figure of merit. Both return `0.0` on the same physical case, from the same numbers. Your literal example is asserted too.

## Blocking 2: the Ainslie Eq. (9.57) entry

Accepted that the entry was not reproducible and had to be rewritten. I reached a different conclusion about which form to ship, and I did not use the physical check you offered, because it does not reproduce. Details below, since this is a permanent public claim and I want the reasoning on the record.

**I kept the true depth H, so the implementation moved rather than the entry.** You framed it as a choice between A (`k²·He²·H/(9·π·η)`, the strict reading) and B (`k²·He³/(9·π·η)`, what the branch shipped), and suggested keeping B with a note. I went with A, for three reasons:

1. **Reproducibility.** An errata entry is a claim that a textbook is defective, so a reader has to be able to check it from the printed page alone. With A the entry is a derivation: take Eq. (9.47) as printed, take Eq. (9.56) as printed, equate at n = 3/2, done. With B the entry has to assert that Ainslie also silently substituted He for H in Eq. (9.47), which is an inference about what the author did, not something the page supports. I would rather the registry contain derivations than reconstructions of intent.
2. **Physics.** Eq. (9.47) comes out of the multipath integral Eq. (9.46), whose `1/(r·H)` prefactor is the cylinder area `A_CS = 2π·r·H` of Eq. (9.44). H there is the depth that counts bottom bounces, and bounces happen off the real seabed. He belongs in Eq. (9.56) because mode angles are set by the apparent pressure-release boundary. The two depths are doing different jobs and the print is right to use different ones.
3. **It fixes the self-inconsistency you flagged rather than documenting it.** `weston_propagation_loss` evaluates Eq. (9.47) with H; now so does the boundary. There is one definition of θ_eff in the module, and a new test, `test_composite_loss_and_the_boundary_use_the_same_effective_angle`, holds them together.

The cost is that the discrepancy is no longer a clean unconditional π. The entry states it as **π·He/H**, and then factors it honestly: the π is unconditional, present even under the He-substituted reading, and the residual He/H is the depth substitution itself, which tends to 1 at high frequency and is presumably where the printed He³ came from. That is more words than "a factor π", but it is what the page actually supports.

**The numbers now match the shipped formula.** For H = 50 m, f = 250 Hz, c = 1500 m/s, sand (η = 0.28 Np/rad, He = 53.628 m): derived 19.9 km, printed 67.1 km, ratio 3.3695 = π·He/H to every digit carried.

**On your independent physical check, I could not reproduce it.** You wrote that at 21.5 km the mode-stripping factor and the peak single-mode factor cross within about 15 %, while at 67.6 km the single-mode branch has overtaken mode stripping by more than a factor of 1.5. With the same inputs, Eq. (9.49) and Eq. (9.54) at `sin² = 1`:

| r | F_MS / F_SM(peak) |
|---|---|
| 19.9 km (derived) | 0.557, i.e. −2.54 dB |
| 21.4 km (He-substituted) | 0.551, i.e. −2.59 dB |
| 67.1 km (printed) | 0.656, i.e. −1.83 dB |

The two branches are equal at 3.6 km and again at 107 km, not near 21 km. And the discrepancy at the printed range (1.8 dB) is *smaller* than at either derived range (2.5 dB), so read literally this test slightly favours the print. Using the depth-averaged rather than peak `sin²` shifts everything by a constant and changes no ordering. I left it out of the entry rather than reword it into something weaker.

**I replaced it with a check that does reproduce**, and that is also independent of how the derivation is read. Eq. (9.56) gives θ₁ = 3.205° and θ₂ = 6.410°, so θ_3/2 = 4.808°. At the derived range Eq. (9.47) gives exactly 4.808°, halfway between the first two modes, which is what the text asks for. At the printed range it has fallen to 2.619°, **below θ₁ itself**. The transition is supposed to mark the stripping of the second mode; at a range where the effective angle is already below the first mode's own angle, the single-mode regime must have begun considerably earlier. That is in the entry, and it is asserted in the test.

**The test now exercises the published rule.** You were right that the old one proved nothing: it wrote `effective_depth` into θ_eff and then asserted the implementation formula back. It now rebuilds θ_eff from Eq. (9.47) with `h`, θ_3/2 from Eq. (9.56) with `effective_depth`, asserts they are equal at the boundary, asserts the closed form, and asserts both the π·He/H ratio to the print and the θ_eff < θ₁ inequality at the printed range.

## 3. NaN bands breaking the advertised chain

Accepted, and fixed at the source rather than documented. Reproduced exactly as you reported (0/1/3/6 NaN bands for the four cases).

A band narrower than the FFT bin spacing holds no energy, and the level of zero energy is `-inf`, not "undefined". So `strike_sel_spectrum` now emits `-inf`, and `weighted_exposure` accepts it, since it is the neutral element of an energy sum. NaN and `+inf` are still rejected. The guides drop the `np.isfinite` mask in all three places (English, Spanish, markdown mirror), and the advertised call now works unaided. The two spectrum renderers map non-finite levels to NaN before plotting so a `-inf` cannot collapse the vertical scale.

Three tests: the chain runs on all four of your cases and gives the same weighted SEL as the masked version, empty bands really are `-inf`, and `+inf` still raises.

## 4. The Southall Table 7 evidence paragraph

Accepted that the paragraph was wrong and had to be rewritten. Two corrections to your account of *why*, one of which changes nothing and one of which does.

**You are right that the +6 dB rule discriminates nothing.** It is satisfied by the printed pair too. Removed.

**On the 15 dB, I think you misread p. 155.** You wrote that the 15 dB is stated as SEL-TTS to SEL-PTS, not SEL to peak, and that it is not in-air specific. The page states two different 15 dB rules, and the entry cited the other one. Verbatim:

> For all marine carnivores in air, there are no published TTS data for impulsive noise exposures. Given the lack of data, a nominal 15 dB offset is used (as in Southall et al., 2007) between the SEL-based TTS threshold and the peak SPL-based threshold.

That is SEL to peak, and it is explicitly in-air only. The one you quoted, "this approach prescribes adding 15 dB to the TTS-onset threshold to estimate PTS onset", is a separate sentence in the next paragraph.

**But your substantive point stands, and is worse than you put it.** That in-air rule does not merely fail to discriminate: it reproduces the *printed* values. PCA 123 + 15 = 138, OCA 146 + 15 = 161, which are exactly the two numbers the errata replaces. So the entry cited a rule that argues against its own conclusion. That needed fixing regardless of who read the sentence how.

**Your +159 dB rule is the right anchor and I have used it.** Evaluating the Table 2 audiogram at the Table 4 f₀ gives PCA −4.56 and OCA 11.41 dB re 20 µPa, hence 154.4 and 170.4 against the corrected 155 and 170. I also ran it across the rows the errata does not touch, which validates the rule before it is applied to the two in dispute:

| group | T(f₀) + 159 | published | residual |
|---|---|---|---|
| SI | 219.59 | 220 | −0.42 |
| PCW | 212.50 | 212 | +0.50 |
| OCW | 226.13 | 226 | +0.13 |
| PCA | 154.44 | **155** | −0.56 |
| OCA | 170.41 | **170** | +0.41 |

HF and VHF are excluded because p. 155 says their peak thresholds "were directly based on empirical data", and LF because the article publishes no audiogram fit for it.

**I added a second line of evidence you did not have: the mechanism.** In each printed row the peak-SPL TTS entry is numerically identical to that same row's PTS-onset *SEL* entry. Reading the four columns in order, printed PCA is 123 / 138 / 138 / 144 and printed OCA is 146 / 161 / 161 / 167. A value duplicated from the neighbouring column, with the +6 dB rule then applied to the duplicate, is the signature of a column slip, and it is exactly the two entries per row that the errata replaces. That explains how the error was made, which the +159 rule alone does not.

**And the entry now records a second, unrepaired inconsistency**: since the errata resolves the in-air peak values in favour of the +159 rule, it supersedes the 15 dB in-air sentence as well as the table, and that sentence is left standing in the article. Worth having in the registry.

## 5. Extending the +159 dB test

Accepted. Parametrised over SI, PCW, OCW, PCA and OCA, with the exclusions above stated in a module constant.

On the tolerance boundary you noticed: the old assertion was `round(threshold) + 159 == approx(212, abs=1.0)`, i.e. 213 against 212 at exactly the limit. The rounding was the problem, not the tolerance. It now asserts on the unrounded threshold, where the largest residual across all five groups is 0.56 dB (PCA), against `abs=0.7`. A companion test asserts that the rule lands more than 8 dB from the printed PCA and OCA values, so the discrimination is pinned and not merely implied.

## 6. The three mutation-coverage gaps

You flagged that these came from reading the tests rather than running mutations. I ran them. All three survived, and a fourth variant I tried survived too. All are now caught except one, which I am leaving deliberately.

| mutation | before | after |
|---|---|---|
| `hearing_range_hz` LF/2018 lower bound 7 → 999 | survives | **caught** |
| NMFS 2024 impulsive HF, offsets broken (193 → 192, 230 → 229) | survives | **caught** |
| NMFS 2024 continuous PW 175/195 → 174/194 | survives | **caught** |
| Southall PCA peak reverted to the printed 138/144 | (n/a) | **caught** |
| NMFS 2024 impulsive HF, whole row shifted by 2 dB with both offsets preserved | survives | still survives |

- **Hearing ranges.** All 24 numbers transcribed from Table ES1 of both documents and verified against the sources first. Southall rows are asserted `None` rather than left unchecked.
- **Impulsive offsets.** `test_southall_impulsive_offsets` is split. The −11 dB SEL step keeps its own test, since it is a Southall-specific worked example, and the +15 dB SEL / +6 dB peak offsets are parametrised over `southall-2019` and `nmfs-2024`. NMFS 2018 publishes no impulsive TTS onset, which its own test already pins.
- **K + C.** `test_weighted_tts_onset_is_k_plus_c` now also asserts `exposure_criteria(g).tts_sel`, so the criteria table and the transcription cannot drift apart. Verified to hold on all 20 rows across the three versions.

The one I left: a whole row of `_CRITERIA_2024_IMPULSIVE` shifted with both offsets intact. Catching that needs a second verbatim transcription of Table A.E-2, which is precisely the duplicate-copies-drifting problem you objected to in the K + C item, and those 48 numbers are already externally verified (by you, against the rendered page). The relations are what the tests should hold; the absolute values are what the source review holds.

## 7. The parabolic-equation bias

Accepted. The deviation is +2.0509 dB (numeric 61.00, flux 58.95), and you are right that the symmetric `abs=3.0` band would have passed a PE reproducing the flux law exactly, which for a 100 m guide at 50 Hz would be wrong. Now `1.5 < numeric - expected < 2.6`, with the docstring saying why the band is one-sided. The modal check keeps `abs=1.0`; the three deviations are −0.29, −0.55 and −0.15 dB.

## 8. Result arrays aliasing the caller's input

Accepted. Both `frequency_hz` and `band_sel` are copied on the way in. `frozen=True` promising immutability while handing back the caller's mutable buffer was indefensible. A test mutates both input arrays after the call and asserts the result is unchanged.

## 9. `>= 0.0` in the exceedance test

Kept the behaviour, documented it, and pinned it. These are onset thresholds and I want the precautionary reading, but you are right that "exceeds" reads as strict and that silence about it was the defect. Both `exceeds_injury` and `exceeds_tts` now state the `margin >= 0` convention, and a test asserts that a peak landing exactly on the criterion reports `peak_margin == 0.0` and `exceeds_injury is True`.

## 10. The v = 2 m/s anchor

No change, and I agree with your reading. At v = 2 the wind term is `22.4·lg(1) = 0`, so the case carries no information about the scaling and is purely the closed form against Ainslie's. It lands at 234.3 m against 241 m, −2.8 % inside a 5 % band. I am keeping it because the three-point set is what the figure caption publishes and dropping the reference point would make the remaining two harder to read, but you are right that it is the one that goes red first if the absorption default moves, and the −0.92 % and +0.07 % of the other two are the ones carrying the scaling.

---

## Also changed while here

- The three underwater guide callouts (English, Spanish, markdown mirror) restated the old "a factor π larger" claim and are rewritten to match the entry.
- The Weston figure alt text said the mode-stripping law runs "to about 21 km"; the boundary moved, so it now says 20 km, in all three places and both languages.
- The `weston_regimes` figure is regenerated in all four variants. Neither new figure uses fills, so no `theme_fill` adoption was needed here.
- The conformance total moved to 463 across 55 domains and 309 standards, with the six hand-written sites updated and `check_conformance_claims.py` green.
- While rebasing I found that the figures gate added in #428 currently fails on `main` itself: the plane-wave band of `duct_mode_cut_on` from #422 is shaded at `alpha=0.10`, which lands 5.3 dE00 from the white page and 4.2 from the black one. That is not this branch's figure, so the one-line fix is in #435 rather than here; until it lands the figures job is red on every branch.

jmrplens:
Follow-up: the SonarCloud analysis of this PR reported seven issues even though the quality gate passed, and those are now fixed too. Five duplicated string literals (the range axis label, the band SEL axis label, the lower-left legend placement, and the two group descriptions shared by the three guidance tables) became named constants; the legend placement went into `_plot/common.py` next to `_LEGEND_UPPER_RIGHT` so it keeps the `Final` literal type that `Axes.legend(loc=...)` needs, which a plain module-level `str` would have broken under mypy. The `_STRINGS` translation table keeps its own literals, since those are lookup keys rather than duplication. In the two exception tests, `float("nan")` and `np.zeros(1024)` are hoisted out of the `pytest.raises` blocks so each block wraps exactly one call that can raise.

SonarCloud now reports zero open issues on the PR.

The branch is rebased onto current `main` (through #425) with the conformance report, api-docs, llms, figures and the six hand-written count sites regenerated at 470 checks across 55 domains and 312 standards.

PR #438, opened 2026-07-30
Orthotropic panel insulation and the limp-frame equivalent fluid

jmrplens:
Two physics additions from the books, both closing a hole in models that were
isotropic-only or rigid-frame-only.

## Orthotropic (ribbed and corrugated) panels

`building/panel_transmission.py`, from Vigran, *Building Acoustics*, Sections
3.7.3.3 and 6.5.3, cross-read against Bies, Hansen & Howard 5e Section 7.2.4.5
(whose Eqs 7.30, 7.31 and 7.38 are an independent transcription of Vigran's
Eqs 6.108, 6.109 and 6.111, which is how I checked mine).

Cladding that is stiff along the corrugations and limp across them has no single
coincidence frequency but a whole range. `orthotropic_critical_frequencies`
returns the pair `(fc1, fc2)` from the two principal bending stiffnesses
(Eq. 6.107), and `orthotropic_transmission_loss` predicts the flattened `R(f)`
over it by either published route:

- `method="integral"` averages the azimuth-dependent transmission coefficient
  of Heckl (1960) and Hansen (1993) over the diffuse field numerically. It is
  the only route that responds to the loss factor. Near-grazing angles are
  excluded either by the size-dependent limiting angle of Bies Eq. 7.36 (pass
  `area`) or by a fixed one (78 degrees by default, Sharp's value, which is
  Vigran's `sin^2 phi = 0.96`).
- `method="heckl"` is Heckl's closed form for `eta = 0`, the design chart of
  Bies Fig. 7.9(b): field-incidence mass law below `fc1/2`, Eq. 7.59 across the
  coincidence range, Eq. 7.60 above `2 fc2`, straight lines in `log10 f` across
  the two gaps. It needs no loss factor and no numerical work, but it requires
  `fc2 > 4 fc1` for its four construction points to stay ordered, and it raises
  a `ValueError` pointing at `method="integral"` when they would not.

`SoundReductionResult` gains `critical_frequency_upper` and `.plot()` shades the
coincidence range, so the orthotropic path composes with the existing Sharp
physical model and with the plateau estimate on one axes and through the same
ISO 717-1 rating. For the sinusoidal profile, `corrugated_plate_stiffness` gives
the Timoshenko and Woinowsky-Krieger equivalent stiffnesses (Eq. 3.115, whose
`Bxz` is the pure twisting term, with no `Bx nu_z` contribution, unlike the
general Eq. 3.114) and `corrugated_plate_mass_factor` the developed-length
increase in surface density that has to accompany them.
`orthotropic_plate_resonance` is the eigenfrequency of a simply supported
orthotropic plate (Eq. 3.113 = Bies Eq. 7.27, after Hearmon 1959), whose lowest
mode bounds the validity of both transmission models.

## Limp-frame equivalent fluid

`limp_frame`, `decoupling_frequency` and `limp_frame_applicable` in
`materials/porous_absorber.py`, from Allard & Atalla, *Propagation of Sound in
Porous Media* 2e Section 11.3.4, Eqs 11.53 to 11.55, after Panneton (2007).

Delany-Bazley, Miki and Johnson-Champoux-Allard all assume a motionless frame,
which only holds above the Zwikker-Kosten decoupling frequency
`Fd = sigma phi^2 / (2 pi rho1)`. Below it a light frame (felts, screens,
aeronautic-grade fibreglass) is dragged along by the pore fluid and its inertia
has to be carried by the equivalent fluid. `limp_frame` corrects the effective
density of any rigid-frame `PorousMediumResult`, keeps its bulk modulus, and
returns another `PorousMediumResult`, so it drops straight into a `PorousLayer`
of the transfer-matrix stack. `limp_frame_applicable` and `LIMP_FRAME_CRITERIA`
carry the published rules of thumb for when a frame may be treated as limp at
all (Beranek's `|Kc/Kf| < 0.05`, the 0.2 to which Doutres et al. 2007 relaxed
it, and the "bulk modulus lower than 20 kPa" that the latter becomes for air).

## Oracles: what is digit-pinned and what is not

**The orthotropic side is digit-anchored.**

- Vigran's own worked example, printed p. 96: a 1 mm steel plate 1 m by 1 m
  gives `f(1,1) = 4.9` Hz and `f(2,2) = 19.7` Hz flat, and `f(1,1) = 25.5` Hz
  and `f(2,2) = 102` Hz once corrugated 20 mm deep at a 100 mm pitch. Four
  published values, and the corrugated pair only comes out with the
  developed-length surface density the same paragraph warns about; with the flat
  7.8 kg/m2 the same formulas return 26.7 Hz and 106.7 Hz. That is what pins the
  mass factor.
- The printed design-chart constants of Bies Fig. 7.9(b) and Eqs 7.59/7.60:
  -54 dB at point A, -13.2 dB across the coincidence branch, -17 dB at point D
  and -23 dB on the recovery branch, all stated for `rho c = 414`. Each is
  recovered by subtracting the leading terms from the computed TL.
- The 25 `h.fc` rows of Hopkins Table A2 (printed p. 608), added to
  `tests/reference_data.py` as the shared source for both the test and the
  conformance report: a multi-row check on the isotropic coincidence frequency
  the orthotropic pair generalises.
- Two exact identities: below `fc1` the double integral collapses to
  `tau = ln((1+q^2)/(1+q^2(1-u)))/q^2`, which it reproduces to 1e-6 dB; and as
  `fc2 -> fc1` the orthotropic impedance becomes Cremer's isotropic one, checked
  against an independently evaluated single integral.

**The limp-frame side is transcription-anchored, not digit-oracled, and it is
worth being exact about what that means.** Allard & Atalla publishes exactly one
table of computed numbers in the whole book (Table 10.2, Rayleigh waves in a
frame in vacuum), and it is not a surface impedance: every
prediction-versus-measurement pair in the porous chapters, Fig. 11.2 included,
is a *figure*. No other source checked (Cox & D'Antonio, Mechel, Atalla & Sgard,
the round-robin and Biot literature) tabulates a limp effective density, surface
impedance or absorption coefficient for a fully specified material.

**What pins the model is the printed Eq. 11.55 itself**, restated term by term
in `test_limp_effective_density_matches_the_printed_closed_form` independently
of the implementation, and behind that test a human reading of the printed page.
The two limits the book states in prose are verified as well, but they are
weaker than their names suggest and it is better to say so than to let them
carry weight they do not bear: **neither constrains the `rho0^2` and `2 rho0`
terms.** A sign-flipped variant of Eq. 11.55 satisfies the heavy-frame limit,
the `omega -> 0` limit *and* the `1/rho1` scaling of the heavy-frame residual,
whose analytic value is `-(rho_eq - rho0)^2 / (rho_t + rho_eq - 2 rho0)`. They
stay in the suite because they would catch a broken composition or a wrong
`rho_t`, not because they identify the form.

Ordered by what each check is actually worth:

- the printed closed form of Eq. 11.55, transcribed in the test: **the anchor**;
- the decoupling frequency on the fully specified glass wool of their Table 6.1
  (printed p. 124), where pure arithmetic gives 43.27 Hz: exact, and independent
  of everything above;
- the printed applicability thresholds 0.05 and 0.2, and the "bulk modulus lower
  than 20 kPa" the second becomes for air;
- the heavy-frame limit stated in prose on printed p. 253 (as `rho1` grows the
  correction vanishes and the rigid-frame result comes back, the residual
  falling as `1/rho1`) and the low-frequency limit (the rigid-frame density
  diverges as `sigma/(j omega)` while the limp one converges on the apparent
  total density `rho_t = rho1 + phi rho0`): corroboration, not identification,
  for the reason above.

Ten new conformance rows, six under the panel section and four under the porous
one. The `limp_frame` docstring, the test module docstring, the CHANGELOG entry
and both guide sections all state plainly that the transcription is the anchor
and that the limits corroborate rather than pin it.

Two notes on the source, for whoever reads the equations next. The pdf-to-printed
page offset of Allard & Atalla is **not constant**: it is +6 in chapter 11 (pdf
259 is printed 253), against +9 in chapter 6 and +7 in chapter 10. And the
paragraph carrying the applicability rules straddles a page break, so the
Beranek 0.05 threshold is on printed p. 253 while the Doutres 0.2 and the
"bulk modulus lower than 20 kPa" it implies for air are on printed p. 254
(pdf 260); the citations now say `pp. 253-254` where they cover both.

## Figures and docs

Two figures via `scripts/generate_graphs.py`, in the usual four variants, both
clearing `check_figures.py` and `check_figure_contrast.py`:

- `orthotropic_transmission_loss`: Vigran's own corrugated geometry against a
  flat sheet of the same material and thickness. Identical within 2 dB below
  `fc1`, then up to 13 dB worse across the coincidence range, `Rw` falling from
  28 dB to 25 dB for a panel that is stiffer and 9 % heavier. The caption is
  explicit that the small low-frequency offset is the extra mass plus the
  different limiting-angle conventions of the two schemes, not a model
  discrepancy.
- `limp_frame_effective_density`: the Table 11.2 material under both models, the
  view of the book's own Fig. 11.2, with the decoupling frequency and `rho_t`
  marked. The two curves merging above the decoupling frequency is the figure's
  own check on the correction.

New sections in the panel-insulation and porous-absorber guides, English and
Spanish, both with the double snippet and the caveats the sources attach (below
`0.7 fc1` the Heckl branch underestimates `R` on small panels; real corrugated
cladding shows a 2 kHz to 4 kHz dip caused by resonances of the panel sections
between the ribs, which neither route predicts; only the sinusoidal profile has
equivalent stiffnesses here, trapezoidal cladding needs its own).

## Gates

`ruff check .`, `mypy src scripts`, `bandit -r src` (0 issues at every
severity), the full `pytest` suite (6884 passed, 69 skipped), `check_figures.py`
(1636 figures), `check_figure_contrast.py`, `check_api_reference.py`,
`check_conformance_claims.py` and regenerated conformance, api-docs, site
reports and llms, all green locally on the rebased head.

Rebased onto `main` at 9499d9ff. The conformance report regenerates to
**484/484 checks across 55 domains and 325 standards** (main is at 474/55/316;
this branch adds ten checks and nine standards), and the six hand-written count
sites are re-fixed from that regenerated report and re-checked with
`scripts/check_conformance_claims.py`: `.zenodo.json`, `docs/getting-started.md`,
both landing-page meta descriptions and the JSON-LD `description` of both
conformance pages.

## Review follow-ups

Two non-blocking points from the review, both fixed:

- `orthotropic_transmission_loss` validated `area` and `limiting_angle` only on
  the integral route, so `method="heckl", area=-5.0, limiting_angle=170.0` was
  accepted silently while the same call raised on the other route. The three
  ignored-by-one-route arguments (`loss_factor`, `area`, `limiting_angle`) are
  now validated together before the method branch, the docstring says which
  route uses each and that all are validated regardless, and a new
  parametrised test pins identical rejection on both routes.
- The CHANGELOG quoted the applicability rule as "softer than about 20 kPa".
  The book prints "a bulk modulus lower than 20 kPa"; the quotation is
  corrected there and in `docs/api-reference.md`, which carried the same
  wording.

## Summary by Sourcery

Add orthotropic panel sound insulation modelling and limp-frame equivalent fluid support, with corresponding API, figures, documentation, tests, and conformance checks.

New Features:
- Introduce orthotropic panel transmission loss prediction, including coincidence range handling and support for both numerical integral and Heckl closed-form methods.
- Add utilities for corrugated panels: developed-length mass factor, equivalent orthotropic bending stiffnesses, and eigenfrequency calculation for simply supported orthotropic plates.
- Provide a limp-frame equivalent fluid model that adjusts rigid-frame porous medium results, along with a Zwikker-Kosten decoupling frequency helper and applicability criteria API.

Enhancements:
- Extend SoundReductionResult to carry an upper coincidence frequency and update plotting to shade orthotropic coincidence ranges.
- Expose new panel and porous-material functions at the top-level API and include them in the generated API reference.
- Generate new example figures for orthotropic transmission loss and limp-frame effective density and wire them into the graph generation pipeline and translations.

Documentation:
- Add English and Spanish guide sections explaining orthotropic panel behaviour and limp-frame porous absorber modelling, including usage examples and caveats.
- Update reference documentation, API reference, changelog, site index, and conformance pages to cover the new orthotropic and limp-frame functionality and revised conformance totals.

Tests:
- Add dedicated test suites for orthotropic panel models and limp-frame porous media, including analytical identities, literature examples, and limit-behaviour checks.
- Expand the numerical conformance report with additional checks for orthotropic panels and limp-frame behaviour, increasing total check and standard counts.

## Summary by CodeRabbit

* **New Features**
  * Added orthotropic/corrugated panel sound-insulation support with coincidence-frequency ranges, resonance calculations, and both integral and Heckl prediction methods.
  * Added limp-frame modeling for porous absorbers, including decoupling-frequency evaluation and applicability checks.
* **Documentation**
  * Expanded English and Spanish guides, API reference entries, worked examples, figure generation, and bibliography updates for the new modeling approaches.
* **Tests**
  * Increased conformance coverage to **484/484** passing checks across **325** standards.


jmrplens (commented):
Independent review against the printed sources (Vigran 2008, Hopkins 2007, Bies/Hansen/Howard 5e, Allard & Atalla 2e), with every oracle recomputed from the books rather than from the branch. The automated reviewers produced nothing actionable on this diff, so this was done as a full adversarial pass: transcription checks digit by digit, independent recomputation of every pinned value, mutation testing of the new oracles, and the house-rules checklist.

## Blocking

Nothing blocking. I looked hard for transcription errors, for a rearranged limp-frame formula that only agrees at the limits, and for an implementation error hiding behind the documented Heckl-versus-integral gap, and found none. Details below so the basis for that conclusion is on the record.

## Non-blocking

1. **CHANGELOG.md (Unreleased, limp-frame entry): misquoted book prose.** The entry puts `the book's "softer than about 20 kPa"` in quotation marks. Allard & Atalla, printed p. 253, actually prints "the limp model is applicable for materials having a bulk modulus lower than 20 kPa". The docstring of `limp_frame_applicable` quotes it correctly; the CHANGELOG (the only shipped text with the wrong wording, the PR body has it too but does not ship) should either quote verbatim or drop the quotation marks.

2. **`orthotropic_transmission_loss`: the `heckl` route skips input validation that the `integral` route enforces.** In `src/phonometry/building/panel_transmission.py` the `method="heckl"` branch returns before the `area` / `limiting_angle` checks, so `orthotropic_transmission_loss(..., method="heckl", area=-5.0, limiting_angle=170.0)` succeeds silently (verified on the branch), while the same call on the integral route raises. Meanwhile `loss_factor` is validated as positive even on the heckl route, where it is equally unused. Either validate all three uniformly before branching or document that the heckl route ignores `area` and `limiting_angle` the way the docstring already documents `loss_factor`. Cosmetic, but the asymmetry will confuse someone.

## Verified correct

**Hopkins Table A2 (printed p. 608, pdf 635):** all 25 `(cL, h.fc)` pairs in `tests/reference_data.py` match the printed table digit for digit, row order included; I also confirmed the identity `h.fc = c0^2 sqrt(12) / (2 pi cL)` is density- and Poisson-independent, so each row is a real check. Worst printed-rounding deviation is 0.048 m.Hz against the 0.06 tolerance, so the margin is thin but correct.

**Vigran worked example (printed p. 96, pdf 118):** recomputed independently from Eqs (3.109)/(3.113)/(3.115) as printed: flat 4.93 / 19.73 Hz, corrugated 25.52 / 102.09 Hz with the developed-length mass (factor 1.0924, which I verified equals both the elliptic closed form and the numeric arc length), and 26.68 / 106.70 Hz with the flat 7.8 kg/m2. The claimed discrimination of `corrugated_plate_mass_factor` is real: the difference (1.15 Hz and 4.6 Hz) dwarfs the 0.05 / 0.1 Hz test tolerances. Eq. (3.115) is transcribed exactly, including the 0.81 and 5/2 constants and the footnoted Blevins misprint warning.

**Bies constants and equations:** Eqs (7.30), (7.31), (7.36), (7.38), (7.58), (7.59), (7.60) and the Fig. 7.9(b) caption (printed pp. 377-384) all match the implementation and the four printed constants -54, -13.2, -17 and -23 dB. The exact values 10 lg(2 pi^2/414) = -13.2167 and 10 lg(2/414) = -23.159 pinned in the tests are correct, and point A reproduces 20 lg(fc1 m) - 54 to 0.12 dB with the 5.5 dB field correction, inside the stated 0.15 dB rounding tolerance.

**The Heckl-versus-integral gap is a book property, not a defect.** At eta -> 0 the Eq. (7.59) branch sits +4.0 to +4.9 dB above the diffuse-field integral across the 500 Hz to 2000 Hz plateau of the Fig. 6.27 case (and the gap grows, not shrinks, if the limiting angle is pushed to 90 degrees, so the 78 degree cutoff is not the cause). Vigran's own Figure 6.27 (printed p. 254) draws Heckl's approximation above its 1 % integral by the same few dB. The integral itself is bracketed on both sides by exact checks that pass: below fc1 it reproduces the elementary mass-law integral to 1e-6 dB, and above 2 fc2 it converges on Eq. (7.60) to under 0.3 dB. Documenting the plateau gap as the accuracy of Heckl's approximation is the right call.

**Limp frame (A&A 11.3.4):** the implementation is the printed Eq. (11.55) verbatim, `(rho_t rho_eq - rho0^2)/(rho_t + rho_eq - 2 rho0)` with `rho_t = rho1 + phi rho0`, not a rearrangement; I checked the printed page directly, and the claimed +6 pdf-to-printed offset of chapter 11 is correct (pdf 259 = printed 253), as are +9 in chapter 6 and the Table 11.2 and Table 6.1 parameter sets. One honesty note the test file already half-makes, stated fully here: the two prose limits alone do **not** pin the `rho0^2` and `2 rho0` terms; a sign-flipped variant satisfies the heavy-frame limit, the omega -> 0 limit and even the 1/rho1 residual scaling (which I confirmed analytically: the residual is `-(rho_eq - rho0)^2 / (rho_t + rho_eq - 2 rho0)`). What pins those terms is the direct transcription test plus a human reading of the printed equation, which this review supplies. The decoupling frequency 43.27 Hz from Eq. (6.90) (printed p. 126) with the Table 6.1 glass wool recomputes to 43.2706 Hz, and the Beranek 0.05 / Doutres 0.2 / 20 kPa thresholds are as printed on p. 253.

**Mutation testing:** six deliberate mutations each killed at least one test: a Hopkins row (12.3 -> 12.5), the 2 pi^2 of the Eq. (7.59) branch, the mass factor forced to 1, the 5/2 constant of Bz, the sign of `2 rho0` in the limp density, and `sqrt(fc1 fc2) -> fc1` in the Eq. (7.60) branch. The oracles are live.

**Composition and house rules:** `orthotropic_transmission_loss` (both methods), `single_panel_transmission_loss` and `plateau_transmission_loss` all return `SoundReductionResult` and all reach `.rating()` through the same ISO 717-1 path; `critical_frequency_upper` is appended last with a `None` default, and the full suite run locally on this head (6844 passed, 69 skipped, no failures) plus the six-platform CI confirm no existing consumer breaks. The ten conformance rows are 6 + 4 as stated, honestly labelled where they are limit-anchors rather than printed digits. EN and ES guide sections are structurally identical and numerically identical token for token (after normalising the Spanish decimal comma in comments). No em-dashes in any added prose, no `assert` in `src/`, no module-level level-2 relative imports in `_plot/`, `docs/api-reference.md` covers all nine new `__all__` names, both new figures come from `generate_graphs.py`, and the new SVGs are vector-only with the generic sans-serif family.

PR #439, opened 2026-07-30
Room-to-room noise reduction and the transmission loss a partition needs

jmrplens:
Every piece of the room-to-room chain was already in the library. The reverberant level a machine builds up in the source room is `steady_state_spl` over `room_constant`, the transmission loss of the partition comes from `building.panel_transmission` or from a measurement, the absorption of the receiving room is `equivalent_absorption_area`, and the verdict is `noise_criterion`. What did not exist is the joint between them, and the joint is the one place where the arithmetic surprises people: **the noise reduction is not the transmission loss of the wall.**

## What this adds

`room_to_room_transmission` walks the chain and returns a `RoomToRoomResult` carrying every intermediate spectrum. The new arithmetic is Norton & Karczub's Eq. (4.101), the steady-state power balance of section 4.9:

```text
NR = TL - 10 lg[ S_w / (S_2 alpha_2 + tau S_w) ]
```

A large partition into a hard room delivers less than its transmission loss; a small one into a well-absorbing room delivers more. The `tau S_w` term is off by default (`include_partition_transmission` switches it on) and `flanking_penalty` records the allowance Norton warns about for mechanical connections and air leaks, so it lands in the table and in the required transmission loss instead of being forgotten.

Everything else is delegation, not reimplementation: the source-room level goes through `steady_state_spl`, the rating through `noise_criterion` / `room_criterion`, the curve sampling through the same helper `duct_path` uses.

`required_transmission_loss` inverts the chain into the number a partition is actually specified from, and `enclosure_required_transmission_loss` does the same inversion for a machine enclosure. `enclosure_insertion_loss` gains `model`: Bies Eq. (7.111) carries a `0.3` inside the logarithm that floors the insertion loss of a fully lined enclosure at `R + 5.2 dB`, Norton Eq. (4.115) has no such floor, and the two diverge once the lining dominates. Reproducing a published answer needs the model its author used.

`steady_state_spl` accepts `distance=None` for the reverberant field alone, the `r -> infinity` limit of Bies Eq. (6.43) and the level a diffuse-field calculation works from, and `source_model` for the three sound power models of Norton Table 4.5. A blower standing in a floor-wall edge modelled as a constant-volume source radiates `10 lg Q = 6 dB` more than its free-space rating, which is exactly what "a conservative estimate" means in problem 4.18 and what separates the printed answer from a number 6 dB low.

## The API decision, and why

The card this came from allowed either a convenience function or documentation-only, with the Norton numbers wired as end-to-end tests over the existing modules. I went with a result object, and I want to argue it rather than assert it.

The case against is real: a wrapper that calls four functions in order adds API surface for little benefit and can drift from the parts it composes. But this is not that wrapper. Three things make the composition more than the sum of the calls:

1. **The middle of the chain is what the practitioner reads.** The band-by-band noise reduction, the receiving room's absorption area, the exceedance over the curve, the transmission loss deficit: a bare chain of four calls throws all of them away and keeps only the last spectrum. The result object is where those live.
2. **The inverse only exists here.** `required_transmission_loss` is Eq. (4.101) solved for `TL` against a criterion curve. It has no home in any of the four parts, and it is the number that goes into a specification.
3. **Consistency with the other composed workflow.** #422 shipped `duct_path`, which answers the same question ("does this room meet NC?") with `rating` / `criterion_curve` / `exceedance` / `meets_target` / `.table()` / `.plot()`. Having one composed workflow be a result object and the other a documentation recipe would be worse than either choice on its own, so this follows `duct_path` field for field.

The drift risk is handled by delegating rather than copying, and by factoring the parts the two workflows genuinely share into `noise_control/_criterion.py` (spectrum coercion, target validation and the four criterion verdicts), which `duct_path` now uses too. The only new arithmetic in this PR is Eq. (4.101) itself and the enclosure inverse.

## The tests, which are the durable part

The three Norton problems ship as end-to-end conformance tests over the real modules, in combination, which nothing did before. Each expected value is a number printed in the book's answer section and each input a number printed in its problem statement.

| Problem | What it pins | Tolerance |
|---|---|---|
| **4.21** (answer p. 614) | Octave-band noise reduction of three partitions into one 8 x 9 x 3 m receiving room: 18.5/26.8/38.0/47.8/47.3/43.9, 36.5/35.8/41.0/48.8/58.3/61.9 and 37.5/40.8/49.0/62.8/65.3/65.9 dB | 0.05 dB (worst 0.031) |
| **4.18** (answer p. 613) | The full chain, blower sound power level to operator-room spectrum: 72.3/60.4/41.4/41.0/33.8/30.7 dB | 0.1 dB (worst 0.068) |
| **4.16** (answer p. 613) | Transmission loss a lined compressor enclosure needs to reach an NC-45 curve: 14.4/25.2/28.9/34.4/35.2/34.7/34.7/31.6 dB | 0.15 dB (worst 0.110) |

All three are registered in the conformance report, together with the exact `10 lg Q` of the constant-volume source model.

## Two things I could not anchor exactly, stated plainly

**The A-weighted row of problem 4.18.** The book prints 56.2/51.5/38.2/41.0/35.0/31.5 dB(A). Weighted with the library's own octave-band corrections (ISO 3744 Annex E Table E.2) the six bands come out within 0.3 dB, not 0.1. Two reasons, both in the source: the book weights with its own Table 4.3, whose 250 Hz value is -8.9 dB against the -8.6 dB of the standard, and its printed 4 kHz answer of 31.5 dB(A) is 0.2 dB below the 30.7 dB level plus the +1.0 dB weighting the same table states. The test pins the unweighted spectrum to 0.1 dB and the A-weighted row to 0.3 dB with that reasoning in the docstring.

**The interior surface split of problem 4.16.** The problem gives the enclosure and machine dimensions but not which surfaces make up the interior absorbing area. Taking it as the lined walls and roof, plus the strip of floor the machine does not cover, plus the five exposed faces of the machine reproduces all eight printed answers to 0.11 dB; counting the whole enclosure floor instead drifts to 0.24 dB. The reading is stated in the test and in the guide, and the residual is within the book's own one-decimal rounding either way.

The NC-45 column of problem 4.16 is also the book's own: it reads 41 dB at 8 kHz where ANSI/ASA S12.2-2019 Table 1 reads 42 dB, agreeing everywhere else. The test uses the printed column so the oracle stays the published one, and a separate test pins that difference so it cannot be mistaken for a bug.

## Documentation

New guide, English and Spanish, walking problem 4.18 from the blower sound power level to the NC verdict, with problem 4.21 as the comparison that shows why the receiving room matters and problem 4.16 as the enclosure case. One figure, the chain against NC 45 with the transmission loss and the delivered noise reduction crossing on the twin axis.

## Summary by Sourcery

Add a composed room-to-room noise transmission workflow, extend enclosure and steady-state room field utilities to support alternative published models, and wire them into plotting, guides, conformance checks, and API exports.

New Features:
- Introduce room_to_room_transmission and RoomToRoomResult to model room-to-room noise reduction, including NC/RC rating, criterion comparison, transmission loss inversion, tabular and plotted views.
- Add enclosure_required_transmission_loss to compute the panel transmission loss an enclosure needs to achieve a target insertion loss, sharing geometry with the forward enclosure calculation.
- Expose new configuration constants ENCLOSURE_MODELS and SOURCE_POWER_MODELS, and support alternative sound power and enclosure correction models across the API.

Enhancements:
- Refactor shared criterion and band-spectrum handling into a noise_control._criterion helper used by duct_path and room-to-room workflows.
- Extend steady_state_spl to support reverberant-only calculations via distance=None and selectable sound power models, aligning with Norton & Karczub formulations.
- Enhance plotting and graph generation to visualize room-to-room chains, including localized labels and twin-axis representations of transmission loss versus noise reduction.

Documentation:
- Document the new room_to_room API, enclosure_required_transmission_loss, ENCLOSURE_MODELS and SOURCE_POWER_MODELS in the reference, and update the API taxonomy.
- Add English and Spanish guides for the room-to-room chain, including Norton's worked problems and enclosure design usage.
- Update conformance report and site metadata to reflect the additional industrial noise control checks and revised totals.
- Extend noise control section overviews and guide listings to include the room-to-room workflow and its design role.

Tests:
- Add end-to-end tests for the room-to-room chain and enclosure transmission loss against Norton & Karczub problems 4.16, 4.18 and 4.21, plus source power model consistency checks.
- Extend enclosure tests to cover the new Norton model option, the inverse transmission-loss calculation, callable targets, and validation paths.
- Add unit tests for steady_state_spl reverberant-only behaviour, new source power models, and validation of the source_model option.
- Add behavioural tests for the room_to_room workflow covering Equation (4.101), options, criterion handling, and plotting.


jmrplens:
@coderabbitai review

PR #441, opened 2026-07-30
Add the critical-band, equally-contributing and octave-band procedures of ANSI S3.5-1997

jmrplens:
ANSI S3.5-1997 is titled *Methods for the Calculation of the Speech
Intelligibility Index* in the plural, and the plural is literal: it defines
four band procedures. `hearing/sii.py` implemented one of them. This adds the
other three.

| `method=` | Bands | Range | Constants |
| :--- | ---: | :--- | :--- |
| `"critical-band"` | 21 | 100 Hz to 9500 Hz | Table 1 |
| `"equally-contributing"` | 17 | 300 Hz to 6400 Hz | Table 2 |
| `"one-third-octave"` | 18 | 160 Hz to 8000 Hz | Table 3, the default |
| `"octave"` | 6 | 177 Hz to 11314 Hz | Table 4 |

## What actually differs between the procedures

Only the band table and the geometry of the upward spread of masking. The
masking slope is one formula throughout, `Ci = -80 + 0.6 (Bi + 10 lg Wi)` with
`Wi` the band width in hertz, and the rest of the chain (self-speech masking,
disturbance, level distortion, band audibility, importance-weighted sum) is
literally the same code for all four:

- the critical-band and equally-contributing procedures spread the masking from
  the upper limit of the masker band up to the geometric centre of the masked
  band, both quantities being tabulated band limits;
- the one-third-octave procedure states that same geometry in terms of band
  centre frequencies, with the two constants printed rounded (`0.89 fi/fk` and
  `10 lg fi - 6.353`). Those are kept exactly as printed, so the existing
  worked-example digits still reproduce;
- the octave-band procedure has no spread of masking at all: an octave band is
  already wider than the spread being modelled, so its equivalent masking
  spectrum level is the equivalent noise spectrum level itself.

So the implementation factors the band table and two masking arrays
(`10 lg Wi` and the band-separation matrix) into a private `_BandProcedure`
record and runs one shared chain, rather than four near-identical paths. The
one-third-octave numbers are unchanged, bit for bit, and its tests pass
untouched.

Table 2 turns out to be the 300 Hz to 6400 Hz span of Table 1, i.e. critical
bands 3 to 19 with an equal importance in each, so its band limits, speech
spectrum and reference internal noise are derived from Table 1 rather than
transcribed a second time. Table 4 tabulates the speech spectrum level and the
reference internal noise at the same figures Table 3 gives at the same six
centre frequencies, which is the expected result for spectrum (per-hertz)
levels and is now a conformance row.

## API shape, and why

`method=` on the existing entry point, not a function per procedure:

```python
speech_intelligibility_index(
    speech_spectrum, noise_spectrum=None, *, threshold=None,
    method="one-third-octave", band_importance=None,
)
```

Three reasons. It matches how the library already exposes alternative
procedures of one standard (`room_noise(method="tangency")`,
`hvac(method="bies")`, `human_vibration(method="linear")`,
`insulation(method="loudspeaker")`, `random_data(method="reverse_arrangements")`);
the four procedures answer the *same* question with the same result type, so a
separate function per procedure would have produced four identical signatures
and four identical result classes; and it keeps the existing call working with
no change at all, which was a hard requirement. The alternative I rejected was
`sii_critical_band(...)` / `sii_octave(...)` and friends: it reads well in
isolation but forces every caller that sweeps procedures to dispatch by hand,
and `SIIResult` would still have needed the `method` field it now carries.

`band_importance=` is a separate argument rather than another `method=` value,
because it composes with all four procedures: it is how the standard's Annex B
functions for particular speech test materials are applied, and it changes only
the weighting, never the audibility chain.

`sii_procedure(method)` returns a frozen, plottable `SIIProcedure` with that
procedure's band centre frequencies, band limits, band-importance function,
reference internal noise spectrum level and normal-effort standard speech
spectrum level. Its `.plot()` steps `Ii` over the band limits, so the four
overlay on one logarithmic frequency axis and the figure below is four
`.plot(ax=ax)` calls.

`SIIResult` gains `method` and `level_distortion` (the clause 5.7 factor `Li`,
which had no accessor and is now the thing that pins the Annex C.1 erratum,
below). The fiche names the procedure in its standard-basis line and tables
whichever bands were used; it still renders to exactly one page for all four,
in both languages, with and without `verbose=True`.

One deliberate limitation: Tables 1, 2 and 4 are implemented with their
normal-effort speech spectrum column, the one the level-distortion factor
needs. `"raised"`, `"loud"` and `"shout"` stay one-third-octave only and the
other procedures raise a `ValueError` that says so, rather than the library
inventing three columns it does not have digits for.

## Oracles

ASA Working Group S3-79, the committee that maintains ANSI S3.5, ships a
reference implementation `SII.C` with eight official test-input files and their
published results, two per procedure, the `_1` file of each pair exercising an
alternative band-importance function. Six of the eight were unused. All eight
now run, as tests and as conformance rows. `SII.C` compiled unmodified with
gcc, printed at `%.17g`, against this branch:

| Case | Published | `SII.C` | phonometry | Δ |
| :--- | ---: | ---: | ---: | ---: |
| `CB.TST` | 0.273 | 0.27293538084817159 | 0.27293538084817154 | 1 ulp |
| `CB_1.TST` | 0.410 | 0.41047412312344123 | 0.41047412312344123 | 0 |
| `ECB.TST` | 0.278 | 0.27813865500646723 | 0.27813865500646717 | 1 ulp |
| `ECB_1.TST` | 0.410 | 0.41047412312344123 | 0.41047412312344117 | 1 ulp |
| `OCTAVE.TST` | 0.491 | 0.49096250624999993 | 0.49096250624999993 | 0 |
| `OCTAVE_1.TST` | 0.323 | 0.32293749999999999 | 0.32293749999999999 | 0 |
| `TO.TST` | 0.445 | 0.44539100585790048 | 0.44539100585790042 | 1 ulp |
| `TO_1.TST` | 0.438 | 0.43821765397620915 | 0.43821765397620921 | 1 ulp |

Nothing had to be tuned to get there; the disagreements are entirely in the
last representable bit and come from summing the same terms in a different
order.

**Eight cases, seven independent confirmations.** `CB_1.TST` and `ECB_1.TST`
agree to every digit, which looked like a copy-paste in the kit until the band
mapping made it obvious: the equally-contributing bands *are* critical bands 3
to 19, the two alternative importance functions weight the same physical bands,
and the two extra critical bands below 300 Hz sit 80 dB under the speech, so
their upward spread of masking never reaches a weighted band. Both procedures
therefore return the identical 0.4104741231 and the kit publishes 0.410 for
each. Eight passing cases would read as eight confirmations, so this is stated
in the test docstring, in the reference-data provenance note, in the module
docstring and in the changelog rather than left for a reader to notice. The
redundant one still earns its place: it breaks the moment either procedure's
band mapping goes wrong.

`OCTAVE.TST` is the Annex C.1 worked example except for a 7.1 dB HL threshold
in the 8000 Hz band where Annex C.1 has 0, so the Annex C.1 input itself, which
the kit's readme prints in full, is anchored separately. That is what makes the
Table C.1 erratum testable rather than only recorded.

## What the anchoring does not cover

Three limits, stated so nobody overcounts what the table above proves.

**The Annex C.1 SII is the committee code's value, not a published digit.** The
kit's readme prints that example's *input* in full, but the printed Annex C
worksheets are still unobtainable, so there is no published SII to compare
against. `0.5039555062` is what `SII.C` returns for that input. The Table C.1
`Li` cell it pins *is* published (in the working group's errata list, as the
corrected 1.00), so that one assertion rests on a printed digit; the index next
to it does not.

**Eight cases are seven confirmations**, as above.

**The three new procedures are implemented against the committee's reference
code and its published results, not against the normative clause text.** The
full text of ANSI S3.5-1997 could not be obtained. What the constants of
Tables 1, 2 and 4 rest on is `SII.C`, which is the standards committee's own
implementation and therefore excellent provenance, cross-checked where an
independent transcription exists: the first six rows of Table 1 agree
digit-for-digit with a third-party reproduction of that table, Table 2 is
derivable from Table 1 and checked to be so, and Table 4's `Ui` and `Xi` agree
with Table 3's at the six shared centre frequencies, which is the expected
result for spectrum levels and would not hold if either table had been
mistranscribed. The one-third-octave procedure that was already shipping is
anchored the same way and additionally against the Hornsby worksheet.

## Errata

The Annex C.1 correction the working group publishes (Table C.1, row i = 5,
`Li` printed 0.10, should be 1.00) is now pinned by a test rather than only
recorded: Annex C.1 is the octave-band example, so it was not computable
before. Clause 5.7 gives 0.99581 for that row, which prints as the corrected
1.00.

`docs/ERRATA.md` gains one entry: the captions of Tables 1 to 4 list a
"hearing threshold levels," column that none of the four tables has (the
working group's list deletes the phrase). It matters more now that all four
tables are implemented, because the quantity the caption invites you to look
for in the table is the user-supplied `Ti'` of clause 5.5, and the column a
reader would mistake for it is the reference internal noise `Xi`.

## Elsewhere

Twelve new conformance rows, taking the ANSI S3.5 section from 10 to 22 and
the report from 487/328 to 499 conformance checks across 338 standards, a new
figure
`sii_band_procedures` overlaying the four band-importance functions, a new guide
section in English and Spanish on the four procedures and when each is used, and
the guide summaries, coverage blocks, API tables and the perception theory
reference widened accordingly.

The masking-slope generalisation is worth one line in its own right: the
one-third-octave `10 lg fi - 6.353` that the standard prints, and that the
perception theory page has carried since the SII landed, is just
`10 lg Wi` for a band `0.2316 fi` wide. Writing it that way is what let the
three new procedures reuse the existing code instead of copying it.

## Validation

Every new procedure is checked against the reference implementation of ASA
Working Group S3-79, the committee that maintains ANSI S3.5, and its published
test-case results, both obtained independently of this code: `SII.C` was
compiled unmodified with gcc and run on the kit's own `.TST` inputs, and the
expected values are the SII the kit's readme publishes for each. No expected
value in this pull request was produced by running the implementation. Sources
are named in the test docstrings and in `tests/reference_data.py`. The limits of
that anchoring are set out above rather than left implicit.

## Checklist

Ran locally, same as CI:

- [x] `ruff check .`
- [x] `mypy src scripts`
- [x] `bandit -r src`
- [x] `pytest -q` (6922 passed, 69 skipped)

Regenerated:

- [x] `make conformance`, `docs/CONFORMANCE.md` committed (499 checks, 55
      domains, 338 standards), `scripts/check_conformance_claims.py` consistent
      and the six hand-written count sites updated
- [x] `make api-docs`, `make llms`, `make site-reports`
- [x] `make graphs` and `scripts/check_figures.py` (1640 figures),
      `scripts/check_figure_contrast.py` (2258 regions)

Applies to new API:

- [x] No new modules, so `scripts/api_taxonomy.py` and `.github/labeler.yml` are
      unchanged
- [x] `docs/api-reference.md` covers every new `__all__` name
      (`scripts/check_api_reference.py`, 1147 names)
- [x] `SIIProcedure` exposes `.plot()`, drawn by `src/phonometry/_plot/hearing.py`
- [x] `pytest tests/test_package_architecture.py` and
      `pytest tests/test_generate_reports.py`

Always:

      (`site/scripts/check-i18n-parity.mjs`)
- [x] `docs/ERRATA.md` records the Tables 1 to 4 caption erratum

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by Sourcery

Extend the Speech Intelligibility Index implementation to support all four ANSI S3.5-1997 band procedures while keeping the existing one-third-octave API as the default, and update documentation and conformance checks accordingly.

New Features:
- Add support for critical-band, equally-contributing critical-band, and octave-band SII methods alongside the existing one-third-octave procedure, selectable via a method parameter.
- Expose SII band procedure metadata through a new SIIProcedure type and sii_procedure() helper, including plotting of band-importance functions.

Enhancements:
- Refactor the SII calculation into a shared chain parameterised by band tables and masking geometry, and extend SIIResult with method and level-distortion outputs.
- Improve plotting and reporting to work across all SII band procedures, with procedure-aware labels in figures and PDF fiches.

Documentation:
- Update English and Spanish SII guides, API reference, errata, and overview docs to describe and illustrate all four band procedures and their usage.
- Add new figures overlaying band-importance functions for the four SII procedures and adjust documentation counts reflecting the increased conformance coverage.

Tests:
- Add extensive tests and reference data covering all four ANSI S3.5 band procedures, including ASA WG S3-79 official cases and Annex C worked examples.
- Expand the numerical conformance report with new SII-related checks and regenerate derived documentation artifacts.

## Summary by CodeRabbit

- **New Features**
  - Added support for all four ANSI S3.5-1997 Speech Intelligibility Index procedures: critical-band, equally-contributing critical-band, one-third-octave, and octave-band.
  - Added procedure selection, custom band-importance weighting, inspectable procedure data, and plotting support.
  - Reports now reflect the selected procedure.

- **Documentation**
  - Expanded guides, API references, Spanish translations, and errata coverage for the four procedures.

- **Tests**
  - Added comprehensive standard-reference and procedure-specific validation.

- **Conformance**
  - Increased reporting to 499 passing checks across 338 standards.


jmrplens:
Adversarial review of this branch, done against the WG S3-79 kit in my source tray: I compiled `SII.C` unmodified with gcc, re-ran the eight `.TST` cases myself, ran a 354-case bit-level sweep of the one-third-octave path against `main`, and ran a 166-entry mutation campaign over the transcribed constants of Tables 1, 2 and 4. Review status of the bots, for the record: only CodeRabbit actually reviewed this diff (nine minor comments, mostly valid prose-consistency fixes worth triaging). Sourcery declined it as over its 150k-character diff limit, Bito refused it on its trial size limit, and cubic skipped. Everything else that is green (Codecov, Sonar, conformance, CI) is a gate, not a review.

## Blocking

**1. The "bit for bit" claim is false: the refactor perturbs the one-third-octave numbers.**

The PR body says "The one-third-octave numbers are unchanged, bit for bit". They are not. On a sweep of 354 inputs (the four vocal-effort spectra in quiet plus 350 random speech/noise/threshold vectors), the branch differs from `main` in the returned `sii` in 93 cases (up to 10 ulps) and in some field of the result (`masking`, `disturbance`, `band_audibility`) in 253 cases. Even the Table 3 `"shout"` spectrum in quiet changes `masking` and `disturbance` bits.

Root cause: `_third_octave_geometry` (`src/phonometry/hearing/sii.py:313`) pre-folds the printed `-6.353` into `bandwidth_db`, so the clause 5.4 slope at line 619 computes `b + (10 lg f - 6.353)` where `main` computed `(b + 10 lg f) - 6.353`. FP addition is not associative.

Verified fix (I patched and re-ran everything): keep `bandwidth_db = 10.0 * np.log10(f)` for the one-third-octave procedure and apply the `6.353` inside `_equivalent_masking` after the addition, i.e. `slope = -80.0 + 0.6 * (self_masked + bandwidth_db - offset)` with `offset = 6.353` for one-third-octave and `0.0` for the critical-band procedures (subtracting `0.0` is exact, so the critical-band association, which matches `SII.C`'s `beta + 10 lg W`, is untouched; carry the offset as a `_BandProcedure` field rather than branching on the name). With that patch the branch is bit-identical to `main` on all 354 sweep cases and every `SII.C` case stays at the same 0-1 ulp distance. Either apply it, or strike the claim from the PR body; given the fiche depends on this path, apply it.

**2. The transitive-pinning argument does not hold: 62 constants survive gross corruption with the suite green.**

The claim is that a mistranscribed band constant would break the 1-ulp agreement with `SII.C`. I mutated every entry of every transcribed array of Tables 1 and 4 (Table 2 is derived), first by one unit in the last printed digit, then grossly (+100 Hz, +1 dB, +0.01 importance). 64 entries survive the fine mutation and 62 survive the gross one, with `tests/hearing/test_sii.py` fully green; I confirmed a representative gross survivor (`_CRITICAL_INTERNAL_NOISE[10]`, -15.4 corrupted to -14.4) against `tests/hearing/`, `tests/test_conformance_report.py` and `tests/test_result_plots.py` as well: 823 passed.

The unpinned entries:

- `_OCTAVE_EDGES`: all seven, even at +100 Hz. The octave procedure has no spread of masking, so the edges enter no formula at all; nothing pins them.
- `_OCTAVE_CENTERS`: all but 1000 Hz (which the 1 kHz lookup pins by accident of `np.isclose`).
- `_OCTAVE_SPEECH_NORMAL`: rows 250, 500 and 8000 Hz (in `OCTAVE.TST` those bands are either fully masked or have `Li = 1` regardless).
- `_OCTAVE_INTERNAL_NOISE`: rows 250, 500, 2000 and 4000 Hz (`x'` sits under the 70/65/25/1 dB noise everywhere it is not directly asserted).
- `_CRITICAL_CENTERS`: rows 7-21, even at +100 Hz. The nominal centres are display-only; the maths uses the geometric centre of the edges.
- `_CRITICAL_INTERNAL_NOISE`: rows 7-21, even at +1 dB. The 70 dB masker in `CB.TST` band 7 drowns `Xi'` in every band above it.
- `_CRITICAL_SPEECH_NORMAL`: rows 7-10 and 18-21.
- `_CRITICAL_EDGES`: 3700, 4400, 5300, 7700 and 9500 Hz survive +100 Hz; 1080 and 1270 Hz survive +-1 Hz.

To be clear about what this is and is not: I checked the constants themselves against `SII.C`'s `cb()`/`eccb()`/`oct()` tables entry by entry, the printed Table 1 head against the CRAN transcription, and the centres of rows 7-21 against the classical Zwicker critical-band centres; **no actual mistranscription exists**. What is wrong is the coverage claim in the PR body, module docstring and test docstring, and it costs nothing to make it true: extend the Table-4-equals-Table-3 check from 1 kHz to all six centres and both columns (kills seven survivors), extend `test_critical_band_table_head` from six rows to all 21 (the CRAN `critical` data frame carries all of them; kills the centre, Xi, Ui and edge survivors), and optionally add one `SII.C`-derived case with unmasked inputs (flat 50 dB speech, low flat noise) so every band's `Ui` and `Xi` participates in the chain.

## Non-blocking

1. **"Tables 1, 2 and 4 are tabulated for normal vocal effort only" misstates the standard.** The CRAN transcription (`plan/SII.pdf`, `head(critical)`) shows Table 1 prints all four vocal-effort columns (normal/raised/loud/shout). The decision to ship only the normal-effort column is fine and correctly motivated (clause 5.7 needs it; the full text is not on hand to transcribe the rest), but say "implemented with" or "carried for", as the guide's "Not covered" paragraph already does, not "tabulated for". Affected: `docs/speech-intelligibility.md:429`, `site/src/content/docs/guides/speech-intelligibility.mdx:479`, `es/.../speech-intelligibility.mdx:497`, and the `ValueError` text at `src/phonometry/hearing/sii.py:645-650` ("is tabulated for the one-third-octave procedure" implies exclusivity).
2. **The wrong-length error message for the default procedure regressed cosmetically.** `_as_band_vector` now prints the computed exact edges, "142.544 Hz - 8979.7 Hz", where `main` said "160 Hz - 8000 Hz". Prefer the nominal range.
3. **The fiche claim is true but untested.** I rendered all 16 combinations (four procedures x EN/ES x verbose on/off): one page each, procedure named in every one. But `tests/hearing/test_sii_report.py` never passes `method=`; add one non-default-method fiche test so the claim stays true.
4. **The CB_1/ECB_1 justification is imprecise.** "the two extra critical bands below 300 Hz sit 80 dB under the speech" (`tests/hearing/test_sii.py:508`, echoed in `tests/reference_data.py`): the masker in those bands is the 10 dB noise line, not anything 80 dB down. The coincidence holds because their spread arrives at ~1e-19 of the local masking energy, below double resolution (I computed it). The conclusion (seven independent confirmations, not eight) is correct.
5. **`tests/reference_data.py:1254-1268`**: the Table 1 head comment is separated from `ANSIS3_5_CRITICAL_TABLE1_HEAD` by the Annex C.1 block; reorder. Also the comment credits `SII.C` for the head digits, but `SII.C` carries no centre frequencies for the critical bands; the centres rest on the CRAN transcription alone.
6. **`src/phonometry/_plot/hearing.py:222`**: the comment says "Spanish labels" but `_SII_METHOD_LABELS` maps method names to the English labels.
7. CodeRabbit's version-banner comment (`llms-full.txt` v3.3.0 vs the `'3.2.0'` example) pre-exists on `main` and is not this PR's defect; its other prose comments are worth applying.

## Verified correct

- `SII.C` compiled unmodified; all eight published results reproduce (0.273, 0.410, 0.278, 0.410, 0.445, 0.438, 0.491, 0.323) and the library agrees with `SII.C` at `%.17g` to within 1 ulp on all eight with nothing tuned. The alternative-importance path is genuinely exercised: with the default importance the `_1` inputs would give 0.273/0.278/0.445, not 0.410/0.410/0.438. (`TO_1.TST` quirk confirmed: its importance line has 19 entries and `SII.C`'s sequential `fscanf` drops the trailing zero; the transcription matches the first 18.)
- CB_1/ECB_1 as one confirmation: verified at the source level (`SII.C`'s `eccb()` tables are `cb()` rows 3-19; the `.TST` inputs coincide on the shared span) and the library's Table 2 derivation matches.
- The Annex C.1 declared limit is accurate on both halves: the `Li` row i=5 correction (printed 0.10, corrected 1.00) is verbatim in the WG errata list, and 0.5039555062 is committee-code-derived as stated (it also rounds to the 0.504 the CRAN vignette quotes from the standard).
- The three corroborations hold as facts: Table 1's first six rows match the CRAN transcription exactly; Table 2 is derivable from Table 1 (and `SII.C` agrees); Table 4's `Ui`/`Xi` equal Table 3's at all six shared centres (I checked all six by hand; note the shipped row pins only 1 kHz, see blocking item 2).
- The existing one-third-octave tests: only the module docstring of `tests/hearing/test_sii.py` changed; no test function was touched.
- API: `method=` and `band_importance=` are keyword-only additions, the default call keeps its signature and procedure; unknown methods raise; per-procedure length checks raise; `"raised"`/`"loud"`/`"shout"` raise outside one-third-octave rather than mis-computing.
- House rules: no em-dashes in any added prose, no `assert` in `src/`, only level-1 module imports in `_plot/`/`_report/`, the three new `__all__` names are in `docs/api-reference.md`, the figure comes from `generate_graphs.py` with generic sans-serif fonts and the `<details>` double snippet, EN/ES pages match heading for heading, fence for fence and number for number, and the new conformance rows pin the published readme values, not the library against itself.

Nothing here says the numbers are wrong; they are right, and the oracle work is the strongest in the hearing area so far. But the two headline claims (bit-for-bit, and transitively pinned tables) are both false as stated, and both have cheap, verified fixes. Fix those before merge.

jmrplens:
**Both blocking fixes are in #442, not in this pull request.** This merged at `916c85f2`, which was the head before I pushed them, so `main` currently carries the three procedures without either fix. #442 replays them onto `eae7474f` and is verified on that base. The rest of this reply is the point-by-point response to the review.

Thank you for this. Both blocking findings are correct, both were claims I made rather than defects you found in the code, and that distinction is the useful part: the numbers were right and my description of why they were trustworthy was not. I reproduced each finding independently before changing anything, and everything below is what I measured, not what I took from the review.

## Blocking 1: the bit-for-bit claim. Confirmed, fixed, and now enforced.

Reproduced with my own sweep (four standard spectra in quiet plus 350 random speech/noise/threshold vectors), comparing `sii`, `masking`, `disturbance` and `band_audibility` against a `main` checkout:

| | before fix | after fix |
| :--- | ---: | ---: |
| cases where `sii` differs | 71 (max 6 ulps) | **0** |
| cases where any field differs | 273 | **0** |

My counts differ from your 93 and 10 ulps only because our random draws differ; the defect and the root cause are exactly as you diagnosed.

Fixed as you proposed: `bandwidth_db` for the one-third-octave procedure is back to `10 lg f`, and the printed `6.353` is carried as a `bandwidth_offset_db` field on `_BandProcedure` and subtracted after the addition, so the slope evaluates in the order the standard prints it. The critical procedures carry `0.0`, and subtracting zero is exact, so their association still matches `SII.C`'s `beta + 10 lg W`. No name branching.

Re-verified after the fix: bit-identical on all 354 sweep cases, and all eight `SII.C` agreements unchanged at 0 to 1 ulp.

I also added a regression test, because the claim is only worth as much as its guard. It rebuilds the clause 5.4 chain from the slope written out in the standard's own order and asserts **bit** equality with `result.masking`, not a tolerance. A reassociation is a few ulps and would pass any reasonable `atol`, which is precisely how this got through the first time.

## Blocking 2: the transitive-pinning argument. Refuted, and closed.

Reproduced before acting. A spot-check of seven representative constants under gross corruption: **five survived** with the suite green. I then ran the full campaign over all 209 entries of every transcribed array, which confirmed the pattern you describe, for the reasons you give: the octave limits enter no formula because that procedure has no spread of masking, the critical-band nominal centres are display-only because the maths uses the geometric centre of the limits, and the 70 dB masker in `CB.TST` drowns `Xi` everywhere above it.

All three of your proposed pins are in:

- **Tables 1 and 4 asserted cell by cell**, all 21 and all 6 rows, every column.
- **Table 4 against Table 3 at all six shared centres and both columns**, replacing the 1 kHz row. You were right that this one mattered most: the limitations section leans on that corroboration, so enforcing a sixth of it was the weakest link in the argument.
- **Twelve flat-input `SII.C` cases**, three regimes per procedure. The quiet regime (0 / -80 dB) makes the disturbance `Xi` in every band; the loud regime (80 / 20 dB) puts the clause 5.7 factor below unity in every band. So each table's `Xi` and `Ui` column respectively moves the answer cell by cell.

Re-running the campaign then left **two survivors**, and they were not mine: `REFERENCE_INTERNAL_NOISE[12]` and `[13]`, Table 3's `Xi` at 2500 and 3150 Hz, which pre-exist on `main`. They survive because the band audibility there is clipped at 1 in every case the suite runs, so a decibel of `Xi` changes nothing downstream. I pinned Table 3 cell by cell as well. It is the one table with a genuinely independent digit source (the Hornsby worksheet), so that pin is a cross-check rather than only a guard.

Final campaign, all 209 entries under gross corruption: **zero survivors**.

Your nuance is in the PR body in your words, not softened: checking the shipped constants against `SII.C`'s own tables, the classical critical-band centres and the independent transcription of Table 1's head found **no actual mistranscription**. The data was right; the argument that the tests protected it was not. The `tests/reference_data.py` comment now labels the Tables 1 and 4 fixtures as transcription pins rather than independent oracles, and states the provenance per column, including that `SII.C` carries no critical-band centre frequencies and those rest on the CRAN transcription alone.

## Non-blocking, all seven

1. **"tabulated for" misstates the standard.** Agreed and fixed in all four places plus the changelog: the standard prints four vocal-effort columns, only the normal-effort one is carried here. "print all four ... but only their normal-effort column is carried here".
2. **Error message regression.** Fixed: it names the nominal centres again, so the default procedure says "160 Hz - 8000 Hz" rather than the computed "142.544 Hz - 8979.7 Hz". Comment added so nobody reintroduces the edges.
3. **Fiche claim untested.** Added: a parametrised test over the three non-default procedures asserting the basis line names the procedure, the table caption matches, and the band column carries that procedure's own first and last centres; plus a Spanish one, since the basis line is translated through a different path.
4. **The CB_1/ECB_1 justification.** You are right and I computed it rather than accepting it: the dropped bands' masker is the input's own 10 dB noise line, and their spread arrives at **1.1e-19** of the local masking energy at the first weighted band. In the lowest *shared* band the same contribution is 5e-2 and does matter, but that band carries zero weight in both alternative functions. Corrected in the test docstring, the reference-data note, the changelog and the PR body. The conclusion stands; the reasoning was wrong.
5. **`reference_data.py` ordering and provenance.** Both fixed. The Annex C.1 block no longer sits between the Table 1 comment and its data, and the comment no longer credits `SII.C` for the centre frequencies.
6. **`_plot/hearing.py` comment.** Fixed: it maps method names to English labels that double as the `_STRINGS` keys.
7. **CodeRabbit's version banner.** Confirmed pre-existing on `main`, left alone. Its prose comments triaged.

## Result

Conformance section 22 to 24 rows, report 499 to 501 checks. Full suite, figure staleness and contrast, api-reference coverage, evidence-page transplant, i18n parity and the changelog zero-deleted-lines property all green; rebased on `098ef6ad`, regenerated rather than hand-merged.

The two headline claims are now either true and guarded, or gone. I would rather have found these myself, and the mutation campaign is the technique I was missing: "would the suite notice if this constant were wrong" is a different question from "do the tests pass", and I had been answering the second while asserting the first.

PR #443, opened 2026-07-30
Run the STIPA and EBU programme oracles in CI from committed data

jmrplens:
Two suites that verify normative criteria were green everywhere by skipping: they only ran on a machine holding a private download. Both now run in CI against committed data.

## Licensing, established first

**stipa.info certified STIPA bench** (Embedded Acoustics BV, Delft). The download page states, verbatim: *"All of the above verification test signals were developed by Embedded Acoustics and may be used for commercial and non-commercial use free of charge."* (<https://www.stipa.info/index.php/download-test-signals>, retrieved 2026-07-30). A broad free-use grant with no redistribution restriction, and the signals exist precisely so that STIPA implementations can be verified against IEC 60268-16 Annexes A and C. Redistributing a small extract for that purpose is within the intent, so: **extract**.

**EBU loudness test set.** *Use of EBU audio test sequences* (v1.0, July 2019) is explicit: *"You may not use any EBU Test Sequences for business, commercial or for-profit activities. You may not copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the EBU Test Sequences."* (<https://tech.ebu.ch/files/live/sites/tech/files/shared/testmaterial/use%20of%20EBU%20AUDIO%20test%20sequences.pdf>, linked from <https://tech.ebu.ch/publications/ebu_loudness_test_set>, retrieved 2026-07-30). The set's own `readme.txt` adds *"The material may only be used for technical testing purposes"* and credits the authentic-programme cases to two feature films whose rights holders gave permission to the EBU, not to downstream users. No file of that set may be committed, so: **derived**, which is what `tests/data/broadcast/` already does.

## What is committed

**`tests/data/stipa/stipa_certified_extract.zip`, 5,789,635 bytes.** The bench is 49 mono 48 kHz WAVs, 133 MB of PCM that does not compress as audio (deflate gets 0.5 %). Each signal is stored losslessly instead: the n-th order difference of its `int16` samples as little-endian `int32`, LZMA-compressed, with the SHA-256 of the original samples in a manifest. Decoding is n cumulative sums and the digest is checked on every load, so what the tests analyse is bit-exact with what stipa.info publishes. 27 of the 49 signals fit in 5.8 MB that way.

| Suite | Committed | Of | Why these |
| :--- | ---: | ---: | :--- |
| C.3.2 direct-method modulation depth | 3 | 11 | `m = 0` (null), `m = 0,9` (worst measured STI deviation, 0,0031) and `m = 1` (top of the dynamic range, worst measured MTF deviation, 0,0065) |
| C.3.3 indirect-method modulation depth | 7 | 7 | complete |
| C.4.2 filter-bank slope | 14 | 14 | complete |
| A.2.2 weighting/redundancy factors | 2 | 6 | the lowest and the highest-STI band pair |
| A.3.1.2 filter-bank phase distortion | 1 | 11 | `TI = 0,9`, the worst measured phase bias (-0,0029) inside the normative `TI = 0,1 .. 0,9` range |

`make_extract.py` next to it rebuilds the archive from a local copy.

**`tests/data/broadcast/ebu_programme_block_loudness.npz`** is unchanged (36 kB, already committed); what changed is that `test_ebu_material_oracle.py` now reads it as its primary path, absorbing the former `test_ebu_material_series.py`.

Total added: **+5.75 MiB** of git objects (`git count-objects -vH` loose size, 154.12 -> 159.87 MiB), essentially all of it the extract.

## What the committed oracles cannot assert

Stated in both module docstrings and both data READMEs rather than papered over.

- **STIPA.** C.3.3 and C.4.2 are complete, so nothing is lost there. C.3.2, A.2.2 and A.3.1.2 keep their ends and their worst case but lose their intermediate points: the shape of the m-to-STI staircase between `m = 0` and `m = 1`, four of the six weighting-factor band pairs, and the phase bias below `TI = 0,9`. Those still run wherever the full bench is present. **No tolerance is relaxed in either mode** — the committed points are asserted at exactly the tolerances the full bench was asserted at.
- **EBU.** The committed series drive the gating and loudness-range stages and must reproduce the published targets (-23,0 LUFS; LRA 5 LU NLR, 15 LU WLR). They cannot cover the stages upstream of the block envelope *on authentic programme material*: 24-bit WAV decoding, the BS.1770 K-weighting front end, the 400 ms/100 ms segmentation and the channel weighting. Those stages are covered in CI by the synthesizable EBU cases in `test_program_loudness.py`; only their behaviour on real programme material is local-only. Where the audio is available the module now also re-measures the envelopes and checks them against the committed file, so the derived oracle cannot drift away from the material it stands for.

## Local vs CI, one convention

Non-redistributable material moves out of `plan/` into the gitignored `tests/data-local/`. Every suite resolves the same way through the new `tests/oracle_data.py`:

1. the dataset's environment override (`STIPA_VERIFICATION_DATA`, `EBU_LOUDNESS_TEST_SET`, `NORAH2_DATA`);
2. `tests/data-local/<dataset>/`;
3. the committed data under `tests/data/` — the path CI takes, which never skips.

`tests/aircraft/test_rotorcraft_norah2.py` moves onto the same resolver (it already committed a 676 kB extract and needed nothing else). `pytest` now prints which copy each dataset resolved to in its run header, so a green run says which oracle produced it:

```
oracle data:
  stipa-verification: committed oracle (27 of 49 bench signals, tests/data/stipa/)
  ebu-loudness-test-set: committed oracle (block-loudness series, tests/data/broadcast/)
  norah2: committed oracle (44-file ARP subset, tests/data/norah2/)
```

`tests/data/README.md` writes the convention down: derived preferred over extract, extract over the full set, four-section README required, the MIT licence does not cover third-party data, removal policy mandatory.

## Verification

Run with `tests/data-local/` moved away, i.e. exactly what CI sees:

```
168 passed, 7 skipped
```

The 7 skips are the full-bench inventory guard and the six audio-path EBU cases, all of which exist only to say the fuller oracle is absent. Both suites themselves execute. With the full sets present the same command gives 175 passed, 0 skipped. Full suite: 6946 passed, 15 skipped. `ruff check .`, `mypy src scripts`, `bandit -r src`, the conformance report and `check_conformance_claims.py` (499 checks) are all clean and unchanged.

## Summary by Sourcery

Run the STIPA and EBU programme oracle suites in CI from committed test data, while unifying how heavy third‑party datasets are resolved between local runs and CI.

New Features:
- Add a committed, lossless extract of the STIPA certified verification bench and wire the STIPA test suite to use it as a CI-safe oracle, preferring a full local copy when available.
- Use the committed EBU programme block-loudness measurement series as a derived oracle so the EBU loudness gating and range tests run everywhere, with optional end-to-end checks from the original audio when present.
- Introduce a shared oracle-data resolver that standardises how tests locate full local datasets versus committed subsets, and surface the chosen source in the pytest run header.

Enhancements:
- Refactor STIPA certification tests to dynamically select available verification signals, decode them from the new extract when the full bench is absent, and add inventory guards for both the extract and full set.
- Extend the EBU loudness oracle tests to validate results both from committed block-loudness series and, where possible, from the full programme audio, including consistency checks between the two.
- Align the NORAH2 rotorcraft tests with the new oracle-data resolution scheme so they transparently prefer a local full release while keeping CI coverage from the committed subset.
- Document the committed-oracle convention, licensing constraints, and removal policies for each dataset under tests/data/, and update contributor and changelog docs to describe the new testing and data-handling behaviour.

CI:
- Clarify in the CI workflow that heavy third-party oracle datasets are never fetched, and that tests rely on the committed derived series and extracts which are guaranteed to run without skipping.

Tests:
- Add tests that verify the integrity and coverage of the committed STIPA extract against its manifest and ensure the certified bench inventory is correct when a full local copy is present.
- Add tests that compare committed EBU block-loudness series against re-measured envelopes from the original audio, ensuring the derived oracle remains in sync with the source material.

## Summary by CodeRabbit

* **Tests**
  * CI now primarily uses committed derived/extracted oracle data for reliable loudness and STIPA/NORAH2 validation.
  * EBU tests always validate against committed series, with optional audio “full chain” checks when local data is available.
  * Added committed-extract integrity/inventory verification and consistency checks (including envelope match for EBU).
  * NORAH2 oracle handling is now resolved via the shared dataset-selection logic.

* **Documentation**
  * Updated dataset-resolution, provenance, and licensing guidance for committed vs local oracle data.
  * Expanded STIPA and broadcast EBU dataset READMEs, plus contributor instructions.

* **Chores**
  * Ignored local, non-redistributable oracle datasets used for full offline test runs.
  * Updated changelog to reflect the CI-wide oracle data changes.


jmrplens:
Thanks, all five picked up.

Applied as suggested: the MD022 blank lines around the new CONTRIBUTING heading, the `text` language on the run-header fence, and the 48 kHz guard in `make_extract.py` (it now fails at generation time rather than at decode time).

The inventory guard was the good catch. Counting per suite did let a same-suite substitution through, so it now pins the exact 27 names via a `_COMMITTED` frozenset built from the same tables the parametrization uses; swapping `m=0.9` for `m=0.1` fails it. While in there I also gave every archive member a fixed timestamp, so rebuilding the extract from the same source is byte-reproducible (verified: two consecutive rebuilds give the same SHA-256).

On the redistribution point, one correction and one concession. The web result your analysis quoted ("not intended for unauthorized redistribution") is a paraphrase, not text from stipa.info. The page's actual redistribution restriction is in the *following* paragraph and applies to the STIPA *measurement* signals sold by instrument manufacturers, which this PR does not touch: "Bedrock allows unrestricted non-commercial use of their signals, but a paid license is required if you intend to redistribute or promote the signal for use with your own STIPA implementation." About the verification signals the page says only, and without qualification: "All of the above verification test signals were developed by Embedded Acoustics and may be used for commercial and non-commercial use free of charge."

That said, you are right that an unqualified free-use grant is not literally the word "redistribute", so the README no longer leaves the reader to infer it. It now states plainly that the grant is unrestricted and draws its redistribution line explicitly around the measurement signals rather than these, that this extract is committed on that reading, that the reading is not confirmed in writing by Embedded Acoustics, and that the removal policy below is the remedy if it is wrong. Written confirmation is worth having and is on me to ask for; it is not something a code change can supply, and it is the same footing as the other third-party data already in `tests/data/`.

jmrplens:
Second round: three of the four are fixed in 455e730, the fourth is a decision rather than a defect.

**Partial local set treated as a full one** — real bug, thank you. The STIPA bench ships as five separate downloads, so unpacking one of them was enough for `Annex C.3.2` to mark the set complete, at which point all five suites enabled and then went looking for files that were never downloaded. Every dataset now declares all the paths that must exist and an incomplete copy resolves to the committed extract; verified by pointing `STIPA_VERIFICATION_DATA` at a directory holding only `Annex C.3.2`, which now reports `committed oracle (27 of 49 bench signals)` and passes instead of raising `FileNotFoundError`. A file missing from a copy that *did* pass the check now reports what is missing and where, rather than silently substituting the extract, which would hide the gap.

**Non-atomic rebuild** — also real. The generator wrote straight into the committed archive, so a bad source file part way through destroyed the previous good copy. It now builds beside the target and renames only after the zip closes cleanly, with the temporary removed on failure. Byte-reproducibility is unaffected (same SHA-256 across rebuilds).

**"Never skips" overclaimed** — correct, and worth fixing precisely because this PR is about honest reporting. What never skips is the *assertions on committed data*; six EBU cases that read programme audio and the full-bench inventory guard do skip in CI by design. Reworded in all six places, including the CHANGELOG and the workflow comment.

**Redistribution authorization** — not actioned, and I want to be plain about why rather than leave it looking overlooked. This is the maintainer's call and it has been made: no approach to Embedded Acoustics, the extract stays, the position is documented as it is. So the README does not claim a permission we do not have. It states that the terms grant use free of charge for commercial and non-commercial purposes, that they do not use the word "redistribute", that committing the extract is **our reading** of those terms and not a permission granted to this repository, and that the removal policy is the remedy. The licence quotation is now cited to the Internet Archive capture of 2025-09-10 rather than the live page, which is currently defaced.

On the substance, one thing your own analysis surfaced actually cuts the other way. The page conditions redistribution explicitly where its author means to — *"a paid license is required if you intend to redistribute or promote the signal"*, about the manufacturers' measurement signals — and attaches no such condition to the verification signals. That is a contrast in the text, not merely an absence of prohibition. It does not amount to a written licence and the README does not pretend it does.

PR #456, opened 2026-07-30
Add the CNOSSOS-EU railway source emission model (Annex II 2.3 and Appendix G)

jmrplens:
Adds the CNOSSOS-EU railway source, section 2.3 of Annex II to Directive
2002/49/EC with the coefficient database of Appendix G, as the sibling of the
road source. It produces the two equivalent source lines of a track, at 0,5 m
and at 4,0 m above the rail head, in the twenty-four 1/3-octave bands the method
works in and energy-summed into the eight octave bands the propagation stage
consumes.

## What is implemented

`railway_source_power` assembles, for every vehicle of the traffic and both
source heights:

- **rolling noise** (2.3.6 to 2.3.10): rail and wheel roughness energy-summed
  and filtered by the contact patch, then turned into sound power per axle by
  the track, wheel and freight-superstructure transfer functions;
- **impact noise** (2.3.11, 2.3.12) at joints, switches and crossings, scaled by
  the joint density and dropped below the minimum speed and while idling;
- **curve squeal**, **traction noise** and **aerodynamic noise** above 200 km/h
  (2.3.13, 2.3.14);
- **bridge noise** (2.3.18) as a separate omni-directional source at 0,5 m;
- the **source directivity** of 2.3.15 to 2.3.17, and the flow term of 2.3.2 for
  a running train or 2.3.4 for an idling one.

Everything underneath is public, so the roughness conversion, the roughness sum,
each rolling component, the squeal rule, the two directivities and the whole
Appendix G database can be used on their own. `VehicleDescriptor` and
`TrackDescriptor` parse the four- and six-digit codes of Tables [2.3.a] and
[2.3.b]. `RailwayEmissionResult` keeps the per-source breakdown and draws both
heights with `.plot()`.

## Which instrument each table comes from

Annex II is layered and getting the layering wrong is the easiest way to ship a
wrong table, so each shipped table records its instrument next to it:

| Instrument | Effect on the railway source |
| :--- | :--- |
| (EU) 2015/996 | The base text: formulae 2.3.1 to 2.3.18 and Tables G-1 to G-7. |
| Corrigendum, OJ L 5, 10.1.2018 | Replaces the whole of Appendix G, and puts the roughness conversion `f = v/lambda` in metres per second, where the published text read km/h. |
| (EU) 2021/1226 | Replaces Tables G-1b, G-2, G-3a, G-4 and G-7; substitutes a new curve-squeal rule and a new bridge model; rewrites the vertical directivity of source A; corrects three band labels and one pair of values in Table G-5. |

The superseded vertical directivity is kept behind `DirectivityEdition` because
the two forms disagree over the whole lower half space, so comparing with a
pre-2021 study needs it. The wavelength-to-frequency resampling, which the
Directive describes in prose only, is selectable through
`RoughnessInterpolation`, and the default is the reading that reproduces the
Commission's own reference source module.

## How it is verified

Annex II prints no worked example and no check value anywhere in 2.3 or
Appendix G, so the two halves are pinned separately and the documentation says
which is which.

**Tables.** Every shipped Appendix G table is compared for exact equality
against constants in `tests/reference_data.py` that were extracted mechanically
from the Official Journal text of the instrument that enacted it, so the
transcription has two independent paths from the same printed page.

**Equations.** The whole chain is pinned end to end on the emission test
workbook the European Commission published with its reference source module.
Fed the 2015 coefficient database that workbook was computed with, the shipped
model reproduces **all 19 872 of its cases at both source heights to
0,0055 dB**, inside the two decimals the workbook prints; 123 of those cases are
committed under `tests/data/cnossos/` with the catalogue they need, and run in
CI without skipping. The committed subset covers every level of every factor the
full set varies, and a coverage assertion requires each committed coefficient
row to appear in a case where it is not masked by a louder term.

Three of the workbook's twenty-three vehicles are left out, because their rows
record defects of the reference program rather than of the method: one carries
`n` instead of a number in its axle-count field and produces no rolling noise at
all, and two have no traction data, which makes visible an all-zero aerodynamic
spectrum the program energy-sums below 200 km/h. `tests/data/cnossos/README.md`
records that, and also what the agreement can and cannot resolve: a coefficient
twenty decibels below its neighbour does not move the total, so the workbook
constrains the loud rows of the database and not the quiet ones.

Everything the 2021 amendment changed postdates the workbook and is therefore
covered by transcription plus closed-form assertion, not end to end. Section 10
of the guide says so in both languages, together with the two points that are
not anchored on any oracle at all: the choice of interpolation rule, and the
proportional split of traction sound power when a real component sits between
the two model heights.

## Errata

Five confirmed defects of the published text are recorded in `docs/ERRATA.md`,
each read off a rendered page image of the Official Journal rather than
extracted text:

- 2.3.2 gives the roughness conversion speed in km/h where it must be m/s, in
  both places where the conversion appears (corrected in 2018);
- the second table of Table G-1 is headed `Lr,VEH,i` although its columns are
  the rail-roughness classes of Table [2.3.b] (corrected in 2018);
- Table G-5 prints 31,4 and 30,7 dB at 6 350 Hz for the 2 200 kW diesel
  locomotive, about 50 dB below both neighbours (corrected to 81,4 and 80,7 in
  2021);
- Appendix G labels its bands 316, 3 160 and 6 350 Hz and four of its
  wavelengths 120, 12, 3,2 and 1,2 mm; the 2021 amendment fixed the band labels
  but left the wavelength labels standing in Table G-1a;
- the 2015 squeal rule uses open intervals that leave R = 300 m and R = 500 m
  with no excess at all, although the same paragraph scopes the effect to radii
  "below or equal to 500 m" (closed in 2021).

One further candidate, the column codes of the 2018 corrigendum's Table G-3, is
left unregistered and the reason is recorded: the corrigendum is published only
as HTML and no rendering of it could be obtained, and this registry does not
record a claim about a printed symbol that has not been read off the page.

## Also in this pull request

Two figures from `scripts/generate_graphs.py` (the octave-band line power of the
two source heights, and the roughness spectrum sliding along the frequency axis
with speed), bilingual guides, eight conformance rows, the API reference and
`llms` manifests, and a CHANGELOG entry under Unreleased.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/456?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added CNOSSOS-EU railway source emission calculations, including rolling, impact, curve-squeal, traction, aerodynamic, bridge, and directivity components.
  * Added support for two equivalent railway source lines with third-octave and octave-band results, per-component breakdowns, and plotting.
  * Added railway vehicle and track descriptors with reusable modelling utilities.

* **Documentation**
  * Added comprehensive railway emission guides and API reference coverage.
  * Added documented regulatory errata and usage guidance.

* **Tests**
  * Expanded conformance reporting to 517 passing checks, including 8 railway-specific validations.


jmrplens:
The three float-equality comparisons Sonar flagged are gone: `density`, `idle` and `q` each sit behind a validation that has already rejected a negative value, so `<= 0.0` selects exactly the same inputs as `== 0.0` and returns the same `-inf`. Each guard now says so in a comment.

The cognitive complexity of `railway_source_power` is addressed rather than waived. It is split into `_roughness_speed_floor` (the 50 km/h floor of 2.3.2, 30 km/h for a tram, which also removes the nested conditional), `_flow_term` ((2.3.2) for a running train, (2.3.4) for an idling one) and `_vehicle_spectra` (the per-vehicle source spectra, built in the order the original built them so no energy sum is reassociated). Because the summation order is what fixes the last bits, this was checked as bit equality and not as a tolerance: a sweep of 16 760 calls to `railway_source_power` over vehicle types, brake types, axle counts, tram flag, superstructure, traction and aerodynamic data, track roughness classes, transfer functions, joint densities, bridges, squeal excesses and section lengths, ten speeds spanning the 50 km/h floor and the 200 km/h aerodynamic threshold, five receiver-angle pairs, both running conditions, both interpolation rules and both directivity editions returns **byte-identical** `third_octave_line_power`, `line_power`, `total_line_power` and every `components` entry before and after.

The three `pytest.raises` blocks now hold exactly the call under test; the vehicles and the zero-length track are constructed above them.

The branch is rebased onto current `main`, the conformance report and the manifests are regenerated (517 checks, 56 domains, 352 standards), and the local gates are green: ruff, mypy, bandit, 7 296 passed and 22 skipped, figures and figure contrast, conformance claims and the curated API table.

PR #463, opened 2026-07-31
Halve the wall time of the test suite

jmrplens:
## What and why

The full suite had grown to 11:30 of wall time (690 s with 12 workers). Four changes bring it to about 5:55 (354 s) without touching what the tests assert: the report generator suite rendered all 67 example fiches twice (once for PDF assertions, once for WebP) as two single-worker monoliths of 297 s and 201 s, and now renders each fiche exactly once as its own parametrized test; the conformance registry memoizes its deterministic, argument-free checks so each computes once per process; the expensive FDTD, elastic-wave and ECMA reference fixtures are shared across xdist workers through load groups instead of recomputing per worker; and the heaviest modules dispatch first. The remaining floor is the steady-state FDTD physics behind the NTFF fixtures, which the convergence assertions are anchored to.

No test was removed, no signal shortened, no tolerance or oracle changed. Coverage is line-for-line identical to main: 251 files, 97 percent, same missing-line sets. Collected tests go from 7756 to 7822, the exact delta of splitting the two fiche monoliths into one test per fiche plus a registry guard.

## Validation

No new computation.

## Checklist

Ran locally, same as CI:

- [x] `ruff check .`, plus `ruff format --check` on the files you touched
- [x] `mypy src scripts`
- [x] `bandit -r src` (any finding fails, including Low)
- [x] `pytest -q`

Regenerated where this change touches them:

- [x] `make conformance`, with `docs/CONFORMANCE.md` byte-identical (the memoized registry produces the same report)

Always:

## Summary by Sourcery

Halve the test suite wall time by restructuring heavy tests, memoizing deterministic conformance checks, and improving xdist distribution of expensive fixtures and modules.

Enhancements:
- Parametrize report generator tests so each example fiche is rendered once, asserting PDF and WebP outputs together.
- Allow the report generator script to run on a configurable subset of registered examples.
- Memoize deterministic conformance registry checks so each outcome is computed once per process.
- Group heavy FDTD, elastic and ECMA reference tests with xdist load groups to reuse expensive module fixtures per worker.
- Front-load the heaviest test modules in pytest collection to better overlap long-running tests with the rest of the suite.

Documentation:
- Document the reduced test suite wall time and its strategies in the changelog.

Tests:
- Add a registry guard test ensuring report examples are registered uniquely and exercised individually.

## Summary by CodeRabbit

* **Performance**
  * Improved test-suite execution by reusing expensive calculations and fixtures.
  * Prioritized slower tests and coordinated related tests for more efficient parallel execution.
  * Report previews now render only once per example.

* **Tests**
  * Expanded validation to ensure each report example produces a valid, one-page PDF and preview.
  * Preserved existing conformance coverage and test behavior.

* **Documentation**
  * Updated the unreleased changelog with these test-performance improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -897,6 +897,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- The test suite runs in about half the wall time. The report generator
+  tests render each example fiche exactly once and assert its PDF and WebP
+  preview on the same run instead of regenerating the full set twice, the
+  conformance checks compute once per test process through a memoized
+  registry, the expensive FDTD, elastic-wave and ECMA reference fixtures are
+  computed once and shared across xdist workers via load groups, and the
+  heaviest modules dispatch first. No test was removed, no signal shortened
+  and no tolerance or oracle touched; coverage is line-for-line identical.
+
 - The eleven documentation clips still encoded as VP9 are now AV1, which is
   what the other fifteen had already moved to, so all 104 WebM files finally
   share one codec. Their 44 files come down from 20.1 MB to 15.5 MB, 23 %
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -902,7 +902,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   preview on the same run instead of regenerating the full set twice, the
   conformance checks compute once per test process through a memoized
   registry, the expensive FDTD, elastic-wave and ECMA reference fixtures are
-  computed once and shared across xdist workers via load groups, and the
+  computed once on a single worker because their tests now share load groups
+  instead of scattering and rebuilding the fixture on each one, and the
   heaviest modules dispatch first. No test was removed, no signal shortened
   and no tolerance or oracle touched; coverage is line-for-line identical.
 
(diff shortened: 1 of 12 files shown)


jmrplens:
Thanks, both looked at.

The changelog wording was wrong and is fixed in 58f73c32: load groups co-locate the marked tests on one worker, where the module fixture is built once. They do not share objects between processes.

The pyproject finding is a false positive. The tests at `test_fdtd_ntff.py:206` and `test_elastic_fdtd.py:116` are unmarked because they do not consume the expensive fixtures at all: `test_dipole_pattern_matches_two_source_oracle` and `test_rayleigh_wave_speed_matches_characteristic_equation` each build their own simulation inline. Every actual consumer is marked: 5/5 for `monopole_far_field`, 3/3 for the mesh fixtures, 2/2 for `body_waves`, 7/7 for the ECMA calibration reference.

The `test_sti.py` and `test_detailed_prediction.py` module fixtures are deliberately left ungrouped because they are cheap. `stipa_signal(48000, seconds=18.0)` measures 0.45 s and the detailed-prediction fixtures are per-band arithmetic. Grouping them would serialise their modules onto one worker to save under a second, which costs more parallelism than it recovers. Only fixtures in the tens of seconds are worth a group.

PR #464, opened 2026-08-01
Link every page to its source and tidy the site head

jmrplens:
## What and why

Reading Starlight's configuration reference against `site/astro.config.mjs` turned up four things worth fixing.

`editLink` was never configured, so no page offered a path to its source. It is now set to `<repo>/edit/main/site/`, which is the prefix Starlight needs given the docs tree lives under `site/`: it composes the base URL with a `filePath` relative to the Astro root. All 298 editable pages (149 per locale) resolve to a file that exists and is tracked in git, each locale to its own source, with the label arriving translated. The 292 generated API pages opt out through `src/routeData.ts` rather than per-file frontmatter, because those files are written from docstrings and the drift gate would revert any edit made through the web anyway.

The head carried a hand-written SVG favicon link duplicating the one Starlight emits from its `favicon` default, so every document shipped two links to the same file. Only the variants the option cannot express remain, which is the pattern the reference documents: the `.ico` for clients that request it by name and the iOS touch icon.

The web app manifest still described the pre-3.0 octave-band filter bank, contradicting the meta description and the JSON-LD on the same pages.

Finally, Starlight 0.41.5 and Astro 7.1.6. Both carry fixes that apply here: the Starlight patch fixes a regression when `docsSchema()` is extended with a union, which is what `content.config.ts` does, and Astro 7.1.6 fixes concurrent static builds failing to generate the i18n fallback rewrites, which is exactly this site's configuration.

## Validation

No new computation.

## Checklist

Ran locally, same as CI:

- [x] Full site build from the committed state: 591 pages, no errors, no new warnings
- [x] `check-i18n-parity`, `check-sidebar`, `html-validate`, `check:math`, links validator
- [x] `pa11y-ci`: 64/64 URLs, 0 errors

Always:

## Summary by Sourcery

Link all editable documentation pages to their source in the repository, exclude generated API reference from edit links, tidy icon handling in the site head, update the web app manifest description, and upgrade Starlight and Astro to pick up relevant framework fixes.

New Features:
- Add edit links in the footer of all editable documentation pages, per locale, pointing to their source files in the repository.

Bug Fixes:
- Prevent edit links from appearing on generated API reference pages that have no hand-editable source.
- Update the web app manifest description to match the current acoustic measurement toolkit documentation.

Enhancements:
- Clean up the document head by relying on Starlight’s default SVG favicon link and keeping only additional icon variants as recommended by the configuration reference.

Build:
- Bump Starlight and Astro dependencies to versions that fix docs schema extension and i18n fallback rewrite issues in the site’s configuration.

Documentation:
- Record the documentation edit-link behavior, favicon cleanup, and manifest description update in the changelog.

## Summary by CodeRabbit

- **Documentation**
  - Added “Edit this page” links to English and Spanish documentation pages.
  - Excluded generated API reference pages from edit links.
  - Updated the web app description to reflect the broader acoustic measurement toolkit.
  - Refreshed the documentation site’s favicon handling.
- **Bug Fixes**
  - Improved documentation-site metadata and page behavior for a more consistent browsing experience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -897,6 +897,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- Every editable page now links to its source in the footer, in both
+  languages. The generated API reference opts out through the route
+  middleware, since those pages are written from docstrings and a
+  regeneration would discard any edit made through the web. The head also
+  drops a hand-written favicon link that duplicated the one Starlight emits
+  from its own default, and the web app manifest describes the toolkit
+  instead of the octave-band filter bank it grew out of.
+
 - The test suite runs in about half the wall time. The report generator
   tests render each example fiche exactly once and assert its PDF and WebP
   preview on the same run instead of regenerating the full set twice, the
--- a/site/astro.config.mjs
+++ b/site/astro.config.mjs
@@ -459,6 +459,14 @@ export default defineConfig({
       ],
       description: siteDescription,
       lastUpdated: true,
+      // "Edit page" link in each page's footer. The docs live under site/ in
+      // the repository, so the base includes that segment; Starlight appends
+      // the page's own path (src/content/docs/...). The generated API tree
+      // opts out in src/routeData.ts: those files are docstring-derived and a
+      // web edit would be overwritten on the next regeneration.
+      editLink: {
+        baseUrl: `${repositoryUrl}/edit/main/site/`,
+      },
       components: {
         // Per-page structured data (TechArticle / BreadcrumbList) and
         // per-page Twitter card tags, layered on the default head.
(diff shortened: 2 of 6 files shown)


jmrplens:
Good catch, fixed in ad7a69d9.

Astro 7.1.6 pins that optional peer to exactly 7.2.2 while the lockfile resolved 7.2.1. Updating `@astrojs/markdown-remark` alone satisfied Astro but left `@astrojs/mdx@7.0.4` holding a second copy, so the tree carried two markdown pipelines instead of one. Tracking the mdx patch release (7.0.5) as well collapses both onto 7.2.2: `pnpm why @astrojs/markdown-remark` now reports a single version, `pnpm install --frozen-lockfile` is clean, and the site builds with 591 pages and no errors.

I did not add `@astrojs/markdown-remark` to `package.json`, since it is not a direct dependency of the site and declaring it would only widen the surface we have to maintain.

PR #467, opened 2026-08-01
Fan the jitted-kernel test job out across the runner's cores

jmrplens:
## What and why

The `tests (numba-jitted kernel)` job ran `pytest -q` serially and took about 1132 s, barely moving when the rest of the suite was halved, because the two techniques that mattered there (load groups and heavy-first dispatch) are inert without `-n`. The reason it was serial turns out to have been an assumption rather than a measurement: the PR that parallelised the suite left it out because "per-worker JIT recompilation would dominate".

Measurement says otherwise. The library jits exactly one function, `_impulse_kernel_py`, already decorated with `cache=True` and reachable only through `time_weighting(mode="impulse")`. Instrumenting the full suite records 12 calls into it in the entire run. Aggregate first-call cost is 37.5 s cold and 3.1 s warm; aggregate machine-code execution is 0.06 s. Compilation is about 0.4 percent of the job and execution 0.005 percent, and the two modules that reach the kernel take the same time with the JIT on as with it off. Serial to four workers, which is the shape of the GitHub runner, measured 1666 s to 441 s on a quiet machine.

A guard step runs before the suite. It compiles both signatures once, so the workers load that machine code rather than each building their own or racing to write the cache, and it fails the job when the kernel produced no machine code. That second part is the point: this job exists to exercise the compiled kernel, and without the guard it could quietly degrade to the pure-Python path and still report green. The environment variable is declared once at job level so the guard and the suite cannot diverge, and the guard requires it to be exactly `"0"` rather than inferring it, since numba defaults to JIT on while this repository's test default is JIT off. `tests/test_check_jit_kernel.py` asserts the guard's verdict follows the environment, so a guard that always passes cannot slip in.

Rejected with numbers: caching numba's directory between runs buys at most the cold compile, `cache=True` is already set, and narrowing the job's scope would drop coverage under the numpy version numba pins.

## Validation

No new computation. No test removed, no signal shortened, no tolerance or oracle touched. Eight rounds of twelve processes compiling simultaneously against a wiped cache produced no failure and left the cache loadable every time.

## Checklist

Ran locally, same as CI:

- [x] `ruff check .`
- [x] `mypy src scripts` (with the numpy the `quality` job installs)
- [x] `bandit -r src`
- [x] `pytest -q`: 7802 passed, 22 skipped, in both the serial and the parallel shape
- [x] `make conformance`: 533/533, no diff

Always:

## Summary by Sourcery

Parallelize the CI job that exercises the numba-jitted kernel and add a guard to ensure the kernel is actually JIT-compiled before running tests.

New Features:
- Add a scripts/check_jit_kernel.py guard script that verifies the numba impulse kernel compiles and warms its cache before the test suite.
- Introduce tests/test_check_jit_kernel.py to assert the guard script’s behavior matches the environment (numba presence and JIT settings).

Enhancements:
- Update the numba-jitted kernel CI workflow to share a single JIT-enabled environment across steps and run pytest with xdist fan-out instead of serial execution.
- Document the new jitted-kernel CI behavior and performance characteristics in the CHANGELOG.

## Summary by CodeRabbit

- **Performance**
  - Improved automated test execution time through parallel processing while maintaining full test coverage.

- **Reliability**
  - Added validation to confirm numerical acceleration is enabled and compiled successfully before performance tests run.
  - Added coverage for environment-dependent validation outcomes and compilation reporting.

- **Documentation**
  - Updated the unreleased changelog with the new test execution improvements and reported benchmark results.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -311,6 +311,22 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    # Declared once for the whole job so the guard below runs under exactly the
+    # environment it vouches for. Split across the two steps, the guard could
+    # stay green on its own NUMBA_DISABLE_JIT while the suite lost its copy and
+    # ran interpreted from end to end.
+    env:
+      # conftest.py disables JIT by default (coverage); override it here, this
+      # is the only job exercising the numba-compiled kernel.
+      NUMBA_DISABLE_JIT: "0"
+      # Pin the numerical thread pools to one thread each so the per-core xdist
+      # workers do not oversubscribe the CPU with nested BLAS/OpenMP pools,
+      # exactly as the main test job does.
+      OMP_NUM_THREADS: "1"
+      MKL_NUM_THREADS: "1"
+      OPENBLAS_NUM_THREADS: "1"
+      NUMEXPR_NUM_THREADS: "1"
+      VECLIB_MAXIMUM_THREADS: "1"
     steps:
     - uses: actions/checkout@v7
       with:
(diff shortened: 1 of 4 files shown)


jmrplens:
Applied in 62f0e2c8, and it matters more than it looks.

The guard's whole purpose is to name what is wrong with the environment. With `numpy` and the kernel module imported at module level, a broken environment killed the script in an import before either check could run, so the failure mode the guard exists to diagnose would have surfaced as a traceback instead of its own message. The two environment checks now come first and the imports follow them.

Verified both ways: with the JIT off and with the variable unset, the script prints its FAIL line and returns 1 rather than raising.

PR #468, opened 2026-08-01
State the docstring equations in LaTeX

jmrplens:
## What and why

The API reference published equations as code snippets, so `power/sound-power` showed `LWA = 10*lg( sum_k 10^(0,1*(LWk + Ck)) )            (Eq. E.1)` where the page beside it typeset the same formula properly. Docstrings now state their equations in LaTeX, and the generator renders them.

The generator change came first, because it turned out `generate_api_docs.py` passed reST math straight through: the `:math:` roles that already existed in the tree were published as literal markup. It now translates `:math:` to inline maths and `.. math::` to a display block, stashing the formula before the prose escaping runs so nothing can corrupt it, and treating a blank line inside a directive as a separator between equations the way Sphinx does.

134 modules carry the conversion: 375 display equations and 3703 inline formulas. Every base-10 logarithm is written `\log_{10}`, the spelling that needs no key to read, rather than the ISO `\lg` the tree used to mix with it. What stays as code is what was never an equation: the report placeholders like `SII = X`, the verbatim quotation of the DIN 45681 Annex J reference program, and tuple literals.

Two rendering defects surfaced while checking the output, both silent. A display equation emitted at column zero inside a list item closes the list in Markdown, which cost `detailed-prediction` two of the four steps of the ISO 12354-1 chain as list items with no error anywhere. And in a table cell the escaping of `|` produces `\|`, which GFM decodes in prose but not inside maths, so KaTeX would typeset a norm where the standard writes a modulus. Both are fixed and both have a regression test.

## Validation

No new computation. The equations were audited against their originals across six domain sweeps, with a second independent pass re-deriving 768 of them over the twenty highest-risk files: no discrepancy, and no disagreement between a documented formula and its implementation. Mechanical checks ran over the whole population rather than a sample: the 134 files are AST-identical once docstrings are stripped, so the branch cannot change behaviour; the 659 decimal commas keep their values, including four ISO thin-space groupings; no exponent was demoted to a coefficient; every `\tag` sits at the end of its equation.

## Checklist

Ran locally, same as CI:

- [x] `ruff check .`
- [x] `mypy src scripts` (272 files)
- [x] `pytest -q`: 7807 passed, 22 skipped
- [x] `make conformance`: 533/533, no drift

Regenerated where this change touches them:

- [x] `make api-docs`, with `site/src/content/docs/reference/api/` and `site/src/generated/api-sidebar.mjs` committed
- [x] Site build: 599 pages, no KaTeX errors, formulas correct inside GFM parameter tables

Always:

## Summary by Sourcery

Express equations in LaTeX across metrology, materials, building, power, vibration and psychoacoustics modules, and regenerate the associated API documentation, while modestly enriching correlation, intensity, distortion and junction utilities with more informative result structures and plotting support.

Enhancements:
- Refine correlation and time-delay utilities with richer result objects, exposing coefficient-based random error and delay uncertainty metadata.
- Extend intensity and modulation distortion analysis APIs with structured results and plotting helpers for field indicators and harmonic/modulation products.

Documentation:
- Rewrite a large number of scientific and standards-based docstrings to express equations in LaTeX, and regenerate API reference pages so formulas render as proper math rather than code.
- Clarify and expand documentation for building acoustics, sound power (pressure and intensity), porous materials, vibration junction transmission, random-data tests, and psychoacoustic tone audibility, aligning text closely with cited standards.


jmrplens:
Four findings, three real and fixed in 719ff73e.

**Formula E.1 (P2): correct and the most serious of the four.** Shortening that summary line to fit the raw-string prefix turned `EN 12354-4 Formula E.1` into `F. E.1`, which names nothing. Restored, and I checked the rest of the diff for the same shape of loss.

**vibration-sound-power decimals (P3): half right.** The two conversions are prose, not quotation, and this project writes docstring decimals with a point, so `101.3 kPa` and `0.118 dB` are correct. But the same file kept `101,3 kPa` in the constant's comment while the equation above it had already moved to a point, so the file disagreed with itself. Unified.

**marine-mammal-weighting orphan (P3): correct.** Fixed in the source docstring and the page regenerated, not edited.

**Double escaping (P1): not reproducible.** The docstrings hold a single backslash. What the check saw is `repr()`, which renders one backslash as two: printing `insulation.__doc__` gives `:math:`D_{nT} = D + 10 \log_{10}(T/T_0)``, a scan for runs of backslashes finds a maximum run length of 1, and the generated markdown carries `\log_{10}`. The site build renders 0 KaTeX errors, which a doubled command would not.

PR #469, opened 2026-08-01
Open the figures at full size and put the markdown copy within reach

jmrplens:
## What and why

Two features, plus an authoring defect the second one uncovered.

**Figures open at full viewport size.** The gallery plots are the clearest case: a third-octave filter bank drawn at 283 px in the page, or 102 px inside a scrolling table on a phone, is not readable, and the figure is the whole point of those sections. `starlight-image-zoom` supplies the dialog. The figure itself is the control rather than a floating icon, with the plugin's button kept underneath so the focus ring, the accessible name and Enter, Space and Escape all still work. Its own labels are hard-coded English, so a rehype pass and a wrapper substitute the Spanish ones and assert the English text they replace is still there, which turns a future rename into a failed build instead of a silent regression. The component that emits the guides' figures now wraps its own light and dark pair, since the plugin only walks the markdown tree and would otherwise have reached the theory pages alone.

**The published markdown copy has a visible handle.** The site already emitted an `index.md` beside every page and advertised it with `rel="alternate"`, but nothing offered it to a reader. `PageActions.astro` is a split button at the top of the right column: copy the page, or open a menu to copy the link to its markdown, view it, hand it to ChatGPT or Claude, or copy the project's BibTeX. The citation is generated from `CITATION.cff`, `VERSION` and the dated changelog heading, and a build-time assertion keeps the About page's blocks in step with it, so the entry cannot drift from the file that owns it. Both new actions carry what they copy in an attribute, so neither makes a request.

Wiring that up showed the `rel="alternate"` guard had always been wrong: it keyed on the route rather than on whether a file was written, so it pointed at a 404 on the 146 Spanish API pages Starlight serves as English fallback, plus the 404 page. The rule now lives in one module that both the generator and the site import, and a page can only advertise a copy that exists. On a fallback page the control offers the English `index.md` actually being rendered, which is the text on the screen.

**The Spanish guides were showing English figures.** Eleven `ThemeImage` call sites on seven Spanish pages passed the English asset, and the component derives the dark variant but never the language, so those figures rendered English axis labels and legends to Spanish readers in both themes. The `_es` files existed all along; these were one-token authoring slips. Fixed, and the convention is now enforced: the parity check fails on a Spanish page that writes a figure as raw HTML or markdown instead of through the component.

## Validation

No new computation. The eleven language corrections were confirmed by extracting every image reference in the Spanish tree: eleven cross-language references on `main`, none on this branch, and none in the reverse direction. The 68 hand-written tags became 34 component calls with every source, alt text and width tier preserved character for character, and a rendered-text comparison of all 598 pages shows no prose changed.

## Checklist

Ran locally, same as CI:

- [x] Full site build, links validator, `html-validate`, `check:math`
- [x] `check-i18n-parity`, `check-sidebar`, `check-contrast`
- [x] `pa11y-ci`: 64/64 URLs, WCAG2AA
- [x] Sweep of all 599 built pages: 0 dead markdown targets across head links, controls and handoff URLs
- [x] Keyboard paths verified in a browser: Tab, arrows, Home/End, Enter, Space, Escape, focus return

Always:

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added site-wide image zoom with keyboard support and localized accessibility labels.
  * Added page actions for viewing and copying Markdown, sharing with AI assistants, and copying citations.
  * Added automatic Markdown links and synchronized citation metadata.
* **Documentation**
  * Improved light/dark image rendering throughout theory documentation.
  * Updated Spanish figures with localized assets.
* **Bug Fixes**
  * Corrected Markdown link availability and removed invalid landing-page Markdown output.
* **Accessibility**
  * Improved focus handling, keyboard navigation, responsive controls, and Spanish translations.


jmrplens:
Fixed in 82fcc7e8. The alert is correct as a defect, though not reachable as a vulnerability here: the values come from `CITATION.cff`, `VERSION` and the changelog heading, all files in this repository, not from request input.

It was still wrong. The pass escaped the five characters BibTeX treats specially but not the backslash, which starts a command and would swallow what follows it, so a backslash anywhere in the citation metadata would have produced a corrupt `.bib` entry. It now runs first, before the accent pass writes backslashes of its own that have to survive untouched, and the generated entry is unchanged: `author = {Requena-Plens, Jos{\'e} M.}` still renders through biber as before.

jmrplens:
Second pass at this, 5eec93af, and the alert was right to keep firing.

My first fix replaced a backslash with `\textbackslash{}`, which is the correct BibTeX spelling but the wrong shape for this input, and the analysis was still detecting the underlying problem: an escape that has to run before the accent pass, and that afterwards cannot be told apart from the backslashes that pass writes itself.

The values here come from `CITATION.cff`, `VERSION` and the changelog heading, all files in this repository. A backslash in any of them is a mistake, not input to be sanitised into something plausible, so it is now refused with a named error the same way an unknown accent and a non-ASCII leftover already were. The generated entry is unchanged.

PR #529, opened 2026-08-11
Flag the ISO 3382-3:2022 revision everywhere the 2012 edition is cited

jmrplens:
The open-plan office metrics implement ISO 3382-3:2012, and the standard has since been revised as ISO 3382-3:2022. This adds the edition note the site already uses for revised standards (the AES17-2015/2020 pattern) everywhere the 2012 edition is presented: the open-plan acoustics guide (EN and ES), the theory page, the central bibliography, and the docs tree.

The note states what actually changed for users of this module: the 2022 revision removes the privacy distance r_P (implemented and printed on the module's report fiche), adds the comfort distance r_C as a reported quantity, and moves the STI measurement to the indirect IEC 60268-16 method. The distraction distance r_D, the spatial decay rate D2,S and Lp,A,S,4m are numerically identical in both editions, which is why the implementation stays declared as 2012 rather than silently relabelled.

No designation changes: every ":2012" stays ":2012". The llms artifacts are regenerated from the updated pages.

## Summary by Sourcery

Flag the ISO 3382-3:2022 revision wherever the ISO 3382-3:2012 open-plan office standard is cited and clarify which edition the library implements.

Documentation:
- Document that ISO 3382-3:2012 has been superseded by ISO 3382-3:2022 in all relevant English and Spanish open-plan acoustics guides and theory pages, including notes on the implemented edition and key changes in the revision (removal of r_P, addition of r_C, STI method change).

Chores:
- Regenerate LLMS text artifacts to reflect the updated documentation.

### DIFF
--- a/docs/buildings/rooms/open-plan-acoustics.md
+++ b/docs/buildings/rooms/open-plan-acoustics.md
@@ -335,7 +335,8 @@ absorption or where in the room a given table sits.
 
 - International Organization for Standardization. (2012). *Acoustics —
   Measurement of room acoustic parameters — Part 3: Open plan offices*
-  (ISO 3382-3:2012).
+  (ISO 3382-3:2012; since revised as
+  [ISO 3382-3:2022](https://www.iso.org/standard/77437.html)).
   [iso.org catalogue](https://www.iso.org/standard/46520.html).
   The open-plan speech-privacy quantities this page implements.
 - Long, M. (2014). *Architectural acoustics* (2nd ed.). Academic Press.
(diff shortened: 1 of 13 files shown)


jmrplens:
Both suggestions describe deliberate choices, so I am keeping the current shape.

The wording is repeated at each surface on purpose: this follows the site's existing edition-note pattern (the same one already used for AES17-2015/2020, IEC 60268-4, ISO 13472-1/-2, ISO 10534-2, ASTM E2611, IEC 61252 and IEC 61265), where every page that cites the implemented edition carries the full note so it stays self-contained offline and in the docs mirror. Identical wording is also what a future maintenance sweep greps for.

Same for the URL: the structured frontmatter references declare only the edition the library implements, and the revision URL rides inline in the note, exactly as the precedent notes do. Adding ISO 3382-3:2022 to the references block would declare an edition the module does not implement, which is what the note exists to avoid.

PR #530, opened 2026-08-11
Complete the ICASSP citation and pin the citation metadata to the release

jmrplens:
Three related fixes to the citation surfaces.

The ICASSP 2023 reference (Drakopoulos et al.) was missing its second author, Arthur Van Den Broucke, and misspelled the first author's given name; both now match the Crossref record for 10.1109/ICASSP49357.2023.10094887, with pages added.

The paper's summary figures had fallen behind the tree: the conformance suite now stands at 533 checks across 57 domains and 362 standards, the package exposes eighteen public domain namespaces, and the bibliography page groups sources with their catalogue records and verified links rather than listing implemented standards only. The text now says what the repository actually contains.

CITATION.cff gains version and date-released. The site's citation module previously omitted them on purpose so the version could not drift from VERSION; that concern is now enforced instead of avoided: the build-time consistency checks assert that CITATION.cff's version equals VERSION and that date-released is not ahead of the CHANGELOG date for that version, so a release that forgets to bump the citation metadata fails the build. Both failure modes were exercised and then restored.

## Summary by Sourcery

Align citation metadata across the site, paper, and CITATION.cff so releases carry complete, consistent citation information and ICASSP 2023 is correctly referenced.

New Features:
- Add build-time checks that enforce consistency between CITATION.cff, VERSION, and CHANGELOG.md for version and release date.

Bug Fixes:
- Correct the ICASSP 2023 bibliographic entry to include the missing author, fix the first author name spelling, and add page numbers.
- Update the paper text so the reported number of conformance checks, referenced standards, domain namespaces, and bibliography description match the current repository state.

Enhancements:
- Extend the site citation loader to require version and date-released fields in CITATION.cff and centralize CHANGELOG.md heading parsing for reuse.
- Prevent stale citation metadata from affecting the site by keeping the site’s displayed version and year sourced from VERSION and CHANGELOG.md while failing the build when CFF drifts.

## Summary by CodeRabbit

* **Documentation**
  * Updated published verification figures to 533 checks across 362 standards and eighteen domain namespaces.
  * Improved bibliography details and descriptions of catalogue records.
* **Release Metadata**
  * Added version 3.3.0 release information.
  * Added consistency checks to ensure release metadata, version, and dates remain synchronized.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -10,6 +10,8 @@ authors:
 repository-code: "https://github.com/jmrplens/phonometry"
 url: "https://jmrplens.github.io/phonometry/"
 doi: "10.5281/zenodo.21215280"
+version: 3.3.0
+date-released: "2026-07-27"
 license: MIT
 keywords:
   - acoustics
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -72,10 +72,12 @@
 @inproceedings{drakopoulos2023,
   title     = {A {DNN}-based hearing-aid strategy for real-time processing:
                one size fits all},
-  author    = {Drakopoulos, Fotis and Verhulst, Sarah},
+  author    = {Drakopoulos, Fotios and Van Den Broucke, Arthur and Verhulst,
+               Sarah},
   year      = {2023},
   booktitle = {Proceedings of the IEEE International Conference on Acoustics,
                Speech and Signal Processing (ICASSP 2023)},
+  pages     = {1--5},
   doi       = {10.1109/ICASSP49357.2023.10094887}
 }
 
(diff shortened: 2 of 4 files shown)


jmrplens:
On the two review suggestions for citation.mjs:

Lexicographic comparison is correct for these values by construction: both sides are ISO YYYY-MM-DD strings, the changelog side comes from a regex that only matches that shape, and the CFF schema validation enforces the same format for date-released. Parsing to Date objects would reintroduce the timezone pitfalls this comparison avoids, so the string comparison stays.

The newest-first assumption predates this change and only feeds the existing release-year fallback, which kept its exact semantics through the refactor. The new assertions look the version up by name, not by position, so they are order-independent. Sorting the parsed headings would change the behavior of code this PR deliberately did not touch.

The missing domains figure is being added.

PR #531, opened 2026-08-11
Fail the build when an llms shard exceeds its byte budget

jmrplens:
Two of the llms shards have been quietly over their 200 kB budget for weeks: llms-start.txt at 208 kB and llms-buildings-insulation.txt at 204 kB. The generator noticed, printed a note, and exited 0, so nothing ever surfaced. The budget exists because larger files get truncated by the fetch tooling AI assistants use, which silently cuts off everything past the limit.

Two changes, in gate-first order:

The generator now treats an over-budget shard as a build failure. The message names the shard, its real size and the limit, and the failure aborts both CI paths that regenerate the artifacts (the docs workflow invokes the script directly; the llms job goes through make, which stops before the diff check). Exactly 200,000 bytes passes; 200,001 fails. With the gate in place first, the then-current tree failed with both offending shards listed, which is the red proof.

The fix is a topic split, no content trimmed: the six theory mirrors and the About page move out of the start shard into a new llms-theory.txt (173 kB), taking start down to 35 kB, and the insulation ratings pages move into llms-buildings-insulation-ratings.txt (77 kB), taking the insulation shard down to 127 kB. The shard index and full-text file pick up the two new entries, and the moved content is byte-identical to what the old shards carried. All 38 shards now sit under budget.

## Summary by Sourcery

Enforce shard size limits for generated llms artifacts and split oversized content into new shards to bring all existing shards back under budget.

Enhancements:
- Add a shard budget check that fails the llms generation run when any shard exceeds the byte limit, instead of only logging a note.
- Introduce manual shard splits to carve out insulation ratings guides into their own shard while keeping folder overviews intact.
- Move theory pages from the start shard into a dedicated theory shard and adjust navigation and labels accordingly.

## Summary by CodeRabbit

* **Documentation**
  * Added a dedicated guide to insulation ratings, including ISO 717, façade sound insulation, outdoor sound radiation, CTE DB-HR calculations, reports, formulas, examples, and limitations.
  * Reorganized theory documentation into its own reference section for easier navigation.
  * Updated full-text documentation indexes and links to include the new theory and insulation-ratings resources.
  * Improved documentation generation checks to prevent incomplete or oversized reference sections.


jmrplens:
Good catch on the About page, with one correction: the code is right and the PR description was wrong. The split moves only the reference/theory mirrors; the About page stays in the start shard, which still lands at 35 kB, far under budget. Moving About into a shard named theory would misfile it. The description above is now fixed to match the code.

The "finite and positive" sentence is a real inaccuracy, but it predates this PR: the ratings pages moved between shards byte-identical on purpose, so the move stays verifiable. It will be corrected at the source (page, Spanish twin and mirror, then regenerated shards) in a separate change, after checking the exact per-field ranges the validator actually enforces.

On the three generator suggestions: the split-label derivation is two comprehensions over a four-entry module-local dict, so a helper would add indirection without removing real duplication; the missing-route error already names the exact routes that failed to resolve, which is the actionable part when a filename changes; and the budget failure already prints every offending shard with its real byte size and the limit on stderr, which CI surfaces as-is. Keeping the gate script minimal wins over structured output it has no consumer for.

PR #532, opened 2026-08-11
Cite ISO 9613-1 where enclosed-space-absorption already uses it

jmrplens:
The enclosed-space absorption guide leans on ISO 9613-1 in its prose (the air attenuation entry of the total absorption budget) but never declared the standard in its references, and offered no path to the page that hosts the full atmospheric absorption model. Both gaps close here.

The references block gains ISO 9613-1:1993 with a scope note that matches what the module actually does: it carries only the six climate-condition power attenuation coefficients of EN 12354-6 Table 1, which are derived from this standard for the 125 Hz to 8 kHz octave bands, while the full model lives in the outdoor sound propagation guide. The See also section now links there. English page, Spanish twin and the docs mirror all updated, with the llms artifacts regenerated.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/buildings/rooms/enclosed-space-absorption.md
+++ b/docs/buildings/rooms/enclosed-space-absorption.md
@@ -242,6 +242,9 @@ repository. Click the preview to open the PDF:
   $D_{nT} = D + 10\lg(T/T_0)$ and $R' = D + 10\lg(S/A)$.
 - [Conformance report](https://github.com/jmrplens/phonometry/blob/main/docs/CONFORMANCE.md):
   the three Annex E cases these implementations are checked against.
+- [Outdoor Sound Propagation](../../environment/propagation/outdoor-propagation.md): the full ISO 9613-1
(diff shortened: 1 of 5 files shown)


jmrplens:
Both points are structural rather than fixable here. The note length follows the existing scope-note pattern for partially-used standards (the ISO 354 clause note on outdoor-propagation is the exemplar), where the constraint lives next to the designation so the references block stays self-contained. And the MDX/Markdown pairing is the docs mirror: the GitHub tree intentionally carries the same content as the site so each renders standalone, which is why the wording matches by construction.

PR #534, opened 2026-08-11
Add the rotorcraft diffraction insertion loss figure

jmrplens:
The rotorcraft screening documentation shows the section geometry and the per-band adjustment, but never the classical curve: how the diffraction attenuation grows with path length difference. This adds that figure, built from the public diffraction_attenuation API swept over delta for four one-third-octave bands at a realistic single-edge geometry (2.5 m edge over the mean ground plane).

The curve family makes the three regimes visible: the threshold below which the term vanishes (delta = -lambda/20, frequency dependent), the grazing-incidence crossing at 10 Ch lg 3 (4.8 dB where Ch = 1, and 3.0 dB at 63 Hz where the low-frequency factor bites), and the 25 dB cap, which only the 4 kHz band reaches within the realizable range, at delta of about 0.68 m. The sweep stops at 1.5 m because that is what the labelled edge can actually produce; every number in the captions was verified against the API.

Four deterministic SVG variants (light/dark, EN/ES), embedded in the guide with the house pattern (snippet, themed image, caption, details block) in the docs tree and both site locales, with the llms artifacts regenerated.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/534?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added rotorcraft terrain-screening guidance for diffraction attenuation across multiple frequency bands and path differences.
  * Included examples covering grazing incidence, shadowed paths, edge height, and frequency-dependent behavior.
  * Added attenuation plots and documented the \(C_h\) factor and 25 dB cap.
  * Updated English and Spanish documentation with translated labels and annotations.
* **Chores**
  * Added the rotorcraft attenuation figure to documentation generation and validation.


jmrplens:
The alt text notation is being fixed to C_h at the source, and the shard copies will follow through regeneration.

The snippet appearing three times is the site's double-snippet convention: every embedded figure carries its own executable snippet per page and locale, and the snippet gate runs them all, so drift shows up as a CI failure rather than silently. Factoring a shared include would break the convention the rest of the corpus follows.

The literal values in the generator are also the house pattern for the deterministic figure corpus: every generator pins its geometry and axis choices as literals so the SVG output is reproducible byte for byte, and the figure-tolerance check guards regressions. Deriving annotation positions from the curves would make the layout drift with any future API change, which is the opposite of what the corpus wants.

PR #537, opened 2026-08-12
Repair the wave-field clips' annotation collisions and the transmission tube's clock

jmrplens:
The frame-by-frame pass over the fourteen wave-field animation clips (FDTD and elastic, both themes, both languages, posters included) recorded twenty-four defects: eight blockers and sixteen minors, from Spanish annotations running off their panels to a physical clock that jumped backwards.

The clock was the transmission tube's: its hold froze a frame from an earlier instant, so time ran 4.5, 4.6, then back to 3.2 milliseconds for the final two seconds, and the posters were baked from the stale frame. The hold now freezes the last simulated frame, monotonic through the cut, and the poster pins the instant that actually shows the transmitted packet. The other blockers cover the ducting clock rendered black on black over the energy map, the barrier's Spanish insertion loss readouts clipped by the panel edge, the aperture captions overflowing their axes, the plate junction's only unprotected label vanishing into the hot spot, and the coincidence arrow that covered the zero of "10 mm" so the plate read as one millimetre thick.

Every repair was verified on the re-encoded frames at the exact timestamps where the defects lived, with measured clearances, and control frames guard against regressions; the ground effect legend now sits in a strip of its own below the reading band, with the three interference minima untouched and still framed by their predicted nulls. Fifty-six re-rendered variants, all AV1, posters and GIFs refreshed.

## Summary by Sourcery

Tighten annotation layout and timing consistency across wave-field visualisation clips, ensuring labels, legends, and clocks remain readable, correctly framed, and physically meaningful in all variants and languages.

New Features:
- Introduce shared helpers to measure rendered text extents and slide annotations to stay within their panels in any language.

Bug Fixes:
- Correct the transmission tube animation clock so the closing hold follows the final simulation frame while the poster snapshots the physically meaningful verdict frame.
- Prevent Spanish and other longer captions from overflowing axes or colliding with arrows, spines, and energy maps in aperture, barrier, diffusion, slit-absorber, plate-junction, and metadiffuser clips.
- Restore readability of labels that previously vanished into hot spots, ground hatching, or black-on-black regions in ducting, refraction, and pillar-hall scenes.
- Ensure insertion-loss readouts and verdict pills are fully visible and consistently sized across twin panels and translations.

Enhancements:
- Reposition legends, time readouts, and captions in ground-effect, ducting, impedance tube, metadiffuser, and pillar-hall figures to avoid tick labels and preserve clear reading bands.
- Refine halo and pill styling for key labels and captions so they maintain contrast without obscuring underlying data or hatching.
- Align plate and hardware labels more robustly with their geometry, avoiding misleading overlaps such as arrows covering thickness markings.

Chores:
- Update poster snapshot selection so the transmission tube poster pins the separated-packet verdict instant while other clips retain physically consistent stills.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved annotation placement across acoustic visualizations to prevent overlapping, clipped, or unreadable labels.
  * Enhanced contrast for timing readouts, verdicts, and key annotations.
  * Corrected animation timing and final-frame handling for transmission-tube visuals.
  * Refined panel framing, legends, captions, and labels for clearer presentation.
  * Improved Spanish and English zoom-panel layouts and localized annotation positioning.


jmrplens:
This one is refuted by the rendered evidence rather than the coordinates. The label anchor sits at y=1.83 with its alignment below the anchor, so the glyphs hang into the panel, not out of it. The exact clearance was measured on the encoded frames at native resolution during the review pass: the microphone digits carry 8 to 12 pixels of air to the boundary in all four variants and both panels, where before this change they sat at zero. Text clipping never enters into it because the glyphs do not reach the limit in the first place.

PR #539, opened 2026-08-12
Derive the hover and idle sources from the ground hover ring

jmrplens:
The rotorcraft chain could consume a hover ring as data but offered no way to build the derived sources the NORAH2 guidance defines from it. This closes that gap: hover_ring_hemisphere turns a measured in-ground-effect ring (constant directivity in the elevation the text prescribes, with the prototype's bearing mapping available as an explicit option) into a standard source hemisphere, and hover_derived_hemisphere applies the published Table 3 offsets to reach out-of-ground hover, full-rpm idle and reduced-rpm idle, with the offsets overridable per call. Taxi enters as the documented selection rule between its two neighbours; the only remaining exclusion, the 150 m side-line fallback, is stated as such.

Validation runs at three levels: closed-form unit tests on the ring geometry and energy averaging, the real committed hover ring, and a new end-to-end case reproduced against the prototype's out-of-ground-hover output at eight microphones over four ground flow resistivities, matching within 0.45 dB with the prototype's own correction values passed explicitly. That case also settled an ambiguity in the guidance text empirically: the constant-directivity reading lands within half a decibel of the prototype only under its bearing mapping, so the default follows the text and the prototype behaviour stays a parameter.

Reading the source pages closely surfaced three defects in the published guidance, now in the errata registry: the prose and the table disagree by 12 dB on the full-rpm idle baseline, the taxi sentence inverts its "respectively", and the guidance's printed offsets disagree with the correction values shipped in the prototype's own database. The conformance suite grows to 535 checks, and the guide gains a hover ring figure with its Spanish twin.

## Summary by Sourcery

Add support for deriving rotorcraft hover, idle and taxi source hemispheres from measured ground microphone rings, and document and validate this behaviour against NORAH2 guidance and prototype data.

New Features:
- Expose hover_ring_hemisphere to build an in-ground hover RotorcraftHemisphere from a ground-ring measurement with configurable mapping and grid.
- Expose hover_derived_hemisphere to generate out-of-ground hover and idle hemispheres from in-ground hover using either published Table 3 offsets or measured per-database corrections.

Enhancements:
- Integrate hover/idle/taxi source derivation into the rotorcraft noise API, plotting utilities, and Spanish/English guides, including new illustrative figures and examples.
- Extend NORAH2 conformance and ARP test coverage with ring-geometry checks and an end-to-end out-of-ground hover case using the reference database.
- Update errata documentation to record discrepancies between NORAH2 guidance prose, Table 3 and shipped correction values, and reflect the expanded conformance suite counts in site metadata and landing pages.

Documentation:
- Update rotorcraft noise documentation (EN/ES) and API reference to cover hover ring and derived hemisphere functions, taxi selection rules, and the limitations of the no-hover-data fallback.
- Add new hover ring figures (light/dark, EN/ES) showing the ring measurement and derived hover/idle sources, and reference them from the guides.

Tests:
- Add unit tests for hover ring interpolation, constant-φ and bearing mappings, derived hemisphere offsets and validation logic.
- Add integration tests parsing NORAH2 ring-format hemispheres and reproducing the Case 4 out-of-ground hover event across multiple microphones.
- Extend conformance checks to cover hover ring-to-hemisphere derivation and Table 3 offset application.

## Summary by CodeRabbit

* **New Features**
  * Added rotorcraft hover-ring processing to create hemispheres and derive out-of-ground-hover and idle sources.
  * Added configurable source mappings, published offsets, and explicit offset overrides.
  * Exposed new hover hemisphere utilities through the public API.

* **Documentation**
  * Expanded rotorcraft guidance for hover, idle, and taxi operations.
  * Added NORAH2 errata and multilingual documentation updates.
  * Updated conformance totals to 535 checks across 364 standards.

* **Validation**
  * Added two rotorcraft checks, increasing coverage from 12/12 to 14/14.


jmrplens:
On the snippet self-containment comment, the resolution is split. The page's snippets are sequential by design and the snippet gate concatenates and executes them per page, so freqs was already defined by an earlier block on the same page and needed no change there. ring_levels and wheeled however never appeared in any earlier block, so both examples now define representative inputs (a synthetic 31-band hover ring and a skid-gear flag), and the ring example carries its own 31-band frequency vector so the comment stays literally true.

The duplicated JSON-LD object predates this branch (it is present twice on main in both language files); this PR leaves one valid object per file, verified by parsing, and the report generator does not reintroduce it.

PR #542, opened 2026-08-13
Set the corpus's negative numbers with a typographic minus

jmrplens:
The figure corpus wrote its negative numbers two ways: axis ticks carry a real minus sign, while the values quoted in annotations, readouts and legends kept the ASCII hyphen of the source code. This wave closes that gap at the root and hardens the clip layout helpers while it is in the area.

A small formatter now writes the typographic minus for every dynamic reading, literals are corrected in place, and the Spanish translation tables move in step with their emitters commit by commit. Thirty-two static assets regenerate, including seven pulled forward because they share a page with an already-corrected sibling, and the Spanish screening-anatomy axis finally drops its doubled preposition, with the surrounding prose reworded to match. Eleven animation clips re-render with the same treatment; two long-parked marks are also repaired, the metadiffuser's stray depth dot and the time-weighting dial label that a neighbouring fix had left hanging between dials.

The layout hardening measures text after translation instead of before, handles negative rotations, clamps against the panel spine, and renames the measurement helper to say what it returns. All of it is churn-free by proof rather than by claim: the five affected clips re-encode frame-for-frame identical, verified by checksum over every frame of a full re-render, so the commit carries code and tests only.

One visible side effect inherited from an earlier fix: the Spanish variants of the loudness gating and block-versus-exponential clips shift sixteen pixels left, which is those clips picking up the Spanish tick minus correction that landed with their formatter, measured as exactly zero additional movement from this branch.

## Summary by Sourcery

Standardize negative numeric values in figure labels and annotations to use a typographic minus, and harden layout helpers so translated labels remain correctly positioned without altering rendered clips.

New Features:
- Ensure both English and Spanish figure assets and animations consistently render negative values with a typographic minus across building, room, signals, spectral, environment, devices, perception, underwater, and field clips.

Enhancements:
- Introduce a shared formatter that emits U+2212 for signed numeric readouts while preserving surrounding hyphens and padding.
- Refine text layout helpers to work on Matplotlib layout boxes, handle negative rotations correctly, and guard against labels that cannot fit within their panels.
- Extend animation rendering to support a pre-measure hook so label fitting runs after translation but before frame generation.
- Adjust gauge endpoint positioning and several figure captions and annotations for clearer association with their panels and geometry.

Documentation:
- Update Spanish outdoor propagation documentation to describe barrier diffraction insertion loss consistently with the revised figure text.

Tests:
- Add unit tests for the minus-sign formatter to ensure only numeric signs are rewritten and field width is preserved.
- Add unit tests for field label layout helpers to verify behaviour with opposite slopes, level labels, and labels that cannot be made to fit within a panel.

<div id='description'>
<a href="https://bito.ai#summarystart"></a><h3>Summary by Bito</h3><ul><li>Updated various binary image and video assets in the .github/images directory.</li>
<li>The changes consist entirely of binary file updates, likely refreshing animations or visual documentation.</li>
</ul></div>

## Summary by CodeRabbit

* **Improvements**
  * Figure annotations now use consistent typographic minus signs and localized signed-number formatting.
  * Improved label placement and fitting in field visualizations, including clearer handling of oversized labels.
  * Updated several diagram labels and measurement displays for improved readability.
* **Documentation**
  * Clarified Spanish descriptions of diffraction-related insertion loss.
* **Tests**
  * Added coverage for numeric formatting and figure-label positioning.


jmrplens:
Three deliberate choices, so leaving them as they stand.

The stderr note fires only on the clamp path, when a label would otherwise run past the panel spine: that is a layout defect a human needs to see, not routine output, and the whole corpus provably never reaches it today. Gating the one signal the guard exists to emit behind a verbosity flag would defeat it.

The formatter is typed to take a number and a format spec, checked under strict mypy, and every call site feeds it fixed example data from the deterministic figure corpus, so non-finite values cannot reach it in practice; if one ever did, a minus-signed infinity would still be the typographically correct output.

The measure callback stays wired per caller on purpose: five clips use it and each fits a different thing (a rotated label, a footer, a pill against a spine). With this population a shared abstraction would be speculation; the pattern can centralize once more clips adopt it and the common shape is real rather than guessed.

PR #543, opened 2026-08-13
Set the diagrams' mathematics with the composer

jmrplens:
With the canvas able to compose mathematics, this migrates the whole diagram corpus onto it: 743 strings across nine modules move from plain text to composed markup, swept module by module so every diagram ends internally consistent rather than patching only the catalogued lines. Variables and lowercase Greek set in italic, capital Greek upright, descriptive subscripts roman and index subscripts italic, with each Spanish twin updated in the same commit under a parity test that demands identical script structure on both sides.

Two typographic calls were settled against the printed sources during the sweep, both verified on the rendered pages: ECAC Doc 29 prints the start-of-roll subscript in italic, so d_SOR and its deltas follow the source rather than the descriptive-subscript list, and ICAO Annex 16 writes the helicopter speed limits as true symbol-plus-subscript, so VH and VNE compose accordingly.

The geometry never moves: across all twelve hundred regenerated variants the non-text markup is byte-identical, full-corpus regeneration reproduces every file twice over, and the untouched diagrams stay byte-for-byte as they were. A handful of Spanish strings picked up small repairs the sweep exposed along the way, a long-wall tag that clipped its last digit once real spaces stopped collapsing, a quarter-wave fraction the twin had left in prose, and two missing accents.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/543?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<div id='description'>
<a href="https://bito.ai#summarystart"></a><h3>Summary by Bito</h3><ul><li>Updated various binary image and video assets in .github/images/</li>
<li>Replaced Unicode minus signs with standard ASCII hyphens in SVG files to improve compatibility</li>
</ul></div>

## Summary by CodeRabbit

* **Improvements**
  * Improved mathematical notation across aircraft, building, device, environmental, material, perception, signal, underwater, and vibration diagrams.
  * Added clearer formulas, symbols, annotations, measurements, and explanatory captions.
  * Improved rendering of uppercase and lowercase Greek letters, including scripted notation.
  * Preserved existing diagram calculations, values, and behavior.

* **Tests**
  * Added coverage for Greek-letter rendering and multilingual diagram text comparisons.


jmrplens:
The exact-string key is deliberate and fails in the safe direction. The exception grants the Spanish twin one extra prose segment for this specific caption; if a future edit rewords the English side, the lookup stops matching and the parity test fails loudly on that pair, forcing whoever reworded it to re-adjudicate whether the exception still applies. A fuzzier key would keep granting the exemption to a sentence nobody has reviewed, which is the silent failure worth avoiding.

PR #546, opened 2026-08-14
Force @puppeteer/browsers past the extract-zip symlink flaw

jmrplens:
The accessibility audit's browser downloader pulled in extract-zip 2.0.1, which has an unfixed symlink path traversal advisory and no patched release. The dependency is development-only, reaching the tree through pa11y and puppeteer, so nothing published to readers or to the package was ever exposed, but the alert has no other way out: pa11y and pa11y-ci both still pin puppeteer to a line whose core requires that exact browser downloader.

The downloader's own next major dropped extract-zip in favour of the platform's tar and unzip, so a scoped override lifts it there. That major is a breaking release upstream, requiring Node 22 and ESM, which is what the workflow already runs.

Since an override forces a version the dependency tree did not ask for, it was verified rather than assumed: the site builds, and the real accessibility audit passes all 66 URLs three times, once from a clean frozen-lockfile install matching CI, the same result as the last green run on the default branch. The lockfile no longer contains extract-zip at all.

<a href="https://cubic.dev/pr/jmrplens/phonometry/pull/546?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<div id='description'>
<a href="https://bito.ai#summarystart"></a><h3>Summary by Bito</h3><ul><li>Added a pnpm override for '@puppeteer/browsers@<3' to version ^3.1.0 to mitigate a security vulnerability related to symlink handling in older versions.</li>
<li>Updated dependency resolutions in pnpm-lock.yaml to include 'supports-color@10.2.2' for several packages.</li>
<li>Removed several outdated or unused packages from the lockfile, including '@puppeteer/browsers@2.13.2', '@tootallnate/quickjs-emscripten@0.23.0', and '@types/yauzl@2.10.3'.</li>
</ul></div>

## Summary by CodeRabbit

* **Security**
  * Updated browser automation package resolution to address unsafe archive extraction behavior in older versions.
  * Ensures affected installations use a safer version range.

### DIFF
--- a/site/pnpm-lock.yaml
+++ b/site/pnpm-lock.yaml
@@ -10,6 +10,7 @@ overrides:
   nanoid@^3: ^3.3.17
   undici: ^6.28.0
   brace-expansion@^1: ^1.1.18
+  '@puppeteer/browsers@<3': ^3.1.0
 
 importers:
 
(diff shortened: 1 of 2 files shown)


jmrplens:
Keeping the caret. The lockfile pins the exact resolved version, so the range never causes silent drift: it only decides what a deliberate lockfile update is allowed to pick up, and every such update goes through CI with the full accessibility audit that exercises this very package. Narrowing to a tilde would also mean editing the override again to accept the next 3.x, which is the opposite of what this repository does with dependencies, and this override exists precisely to stay ahead of an unpatched advisory.

PR #787, opened 2026-09-15
How a vibration is predicted before it exists, and how a seat and a meter go on the bench (DIN 4150-1, E DIN 45672-3, ISO 4866, ISO 10326-1, ISO 8041-1)

jmrplens:
The same for the predictions and the bench work: five guides that showed what is computed and nothing of what stands in front of the person computing it. Each gains a diagram, and the two bench guides gain a section on how the measurement goes as well, in both languages and in the mirror.

Vibration prediction gains the path DIN 4150-1 lays down for propagation and transfer: the near field inside the far-field boundary where its approximations do not hold, the decay of Formula (2) beyond it with the surface wave and the body waves, and the building on the right sitting on its ground as a mass on a spring and handing the motion on through its foundation to its floors.

Railway prediction gains the chain of E DIN 45672-3 drawn over a street tram line and the building beside it: the emission spectrum taken at its point on the ground, the track bed where a mitigation measure would go, the ground term under its power law, the transfer into the foundation and up to the middle of a floor, where each term of Formula (1) belongs, and the two quantities Clause 7 turns the resulting spectrum into for DIN 4150-2.

Building response gains what a building's own frequency is predicted from under Annex D of ISO 4866: an eighteen-storey building drawn to scale and what each predictor reads off it, the storey rule and Formulae D.1 to D.3, whose coefficient ranges give a span of frequencies on one building rather than one frequency, with the fit of Figure D.1 and the damping of D.4 in two boxes at the foot.

Seat vibration gains the rig of ISO 10326-1: the suspension seat on the platform of the simulator with the test person on it, the mounting disc in section with its dimensions, the platform accelerometer below the seat accelerometer, and the two weighted chains that produce the seat and platform magnitudes. The procedure follows the standard: a chain to ISO 8041 alike at both points, calibrated to ISO 16063-1 before and after each series, the safety requirements of ISO 13090-1, the run-in with a 75 kg inert mass near the natural frequency of the suspension, the ride height of 8.1.3, an adjustable backrest set about 10 degrees back, the platform accelerometer inside a 200 mm circle centred below the seat accelerometer, and three consecutive runs whose seat values lie within 5 % of their mean with the platform inside its input tolerances, once per test person, since Clause 11 asks each of them to pass. The module docstring now says the same thing: 10.2.1 asks for that agreement at the seat, and holds the platform to the input tolerances of 9.2.

Meter verification gains the pattern evaluation bench of ISO 8041-1: the mechanical test of 12.11.2, with the meter's transducer on a calibrated reference transducer on a vibration exciter; the electrical test of 12.11.3 and 12.13, through the maker's substitute for the transducer; the saw-tooth bursts of Table 6; the 60 dB linear operating range the test levels are set in; and a box with Formulae (9), (12) and (10). The procedure follows Clause 12: bench instruments traceably calibrated with every test frequency within 0.2 %, at least three readings at the reference frequency no more than 3 % apart with their mean inside Table 2, Formula (9) within 4.5 % expanded uncertainty with both transducers mounted as ISO 16063-21 describes, Formula (12) within 3 %, the transducer error graded against Table 5, the phase check of Annex H, and the fixed frequencies of 13.10 for a one-off instrument. The periodic verification of Clause 14 is set out after it as the lighter bench it is.

It also re-cites the f = 46/h fit. The module, its conformance rows and the building frequency figure placed the fit and its band in D.3, and neither is where the standard puts them. On printed folio 25 of ISO 4866:2010, PDF page 31, both sentences sit in D.2, the clause on empirical methods: Reference [48] recommends f = 46/h Hz, Figure D.1 shows the fit, and it is that paragraph that says errors of plus or minus 50 % are not uncommon. D.3 is the clause on computer-based methods and says nothing about either, and the numbered formulas D.1 to D.3 are the three code expressions for the period, so 46/h is not Equation (D.3) at all. The curve is now named by the figure it is plotted in and the band by the clause that writes the sentence.

Two figures that already existed come back with that correction and with the subscripts the standards print, now that the canvas sets upright the indicated value and the input of ISO 8041-1 and the seat and platform pair of ISO 10326-1.


jmrplens:
@coderabbitai review

